PR #141, opened 2022-04-27
Upgrade our own version of wireit and enable GitHub caching

aomarks:
Bumps the version of Wireit we use to test Wireit to `0.2.0`, and enables GitHub Actions caching.

The 2nd commit in this PR is a minor README fix. You can see that it finished in only a few seconds, because all of the tests were already cached from the first commit:

https://github.com/google/wireit/runs/6185010301?check_suite_focus=true

image

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,6 +36,8 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: npm
 
+      - uses: google/wireit@setup-github-actions-caching/v1
+
       - run: npm ci
       - run: npm test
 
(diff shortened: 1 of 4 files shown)


AndrewJakubowicz (commented):
~Does the CHANGELOG need updating?~

Edit: I get it!

aomarks:
> Does the CHANGELOG need updating?

I don't think so, most of the changes are just for our repo. The README change technically affects the package, but it's really minor and I think people almost never consume the README through npm anyway.

AndrewJakubowicz (approved):
I get it! AWESOME!

augustjk (approved):
Exciting!

PR #144, opened 2022-04-27
Parse JSON into an AST that preserves locations

rictic:
Propagate those AST nodes and locations into errors, but don't yet use them anywhere

Recommend reviewing util/ast.ts first, as that's the API surface that we use elsewhere


src/event.ts
```diff
@@ -9,6 +9,7 @@ import type {
   ScriptReference,
   PackageReference,
 } from './script.js';
+import {AstNode, ParseError} from './util/ast.js';
```
aomarks, line 12: I've been exclusively using `import type` for types. I like how it visually separates actual runtime module dependencies from type-system-only ones.

I think that's the only benefit that applies to us, though. If we wanted to (partly) enforce it, we could set `"importsNotUsedAsValues": "error"`.

But maybe this style is too inconvenient to follow, WDYT? Maybe would be nice if VSCode did this automatically.

src/script.ts
```diff
@@ -63,7 +66,33 @@ export interface ScriptConfig extends ScriptReference {
    * - "if-file-deleted": If an input file has been deleted, and before restoring from
    *   cache.
    */
-  clean: boolean | 'if-file-deleted';
+  clean: AstNode<boolean | 'if-file-deleted'> | undefined;
```
aomarks, line 69: I like having default values made explicit by the `Analyzer`, so that there's never a mixup over what the default is. `clean` is accessed in two places in `Executor`, so both need the `?? true` now.

Can we preserve that and not add `undefined` here?

src/util/ast.ts
```diff
@@ -0,0 +1,105 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
```
aomarks, line 7: Consider `import * as jsonc from 'jsonc-parser'` so that the similar-sounding things are more clearly distinguished, and we don't need to pick aliases.

src/util/ast.ts
```diff
@@ -0,0 +1,105 @@
...
+ * A safer override, preferring unknown over any.
+ */
+export interface AstNode<T extends ValueTypes = ValueTypes>
+  extends AstNodeInternal {
+  value: T;
+  children?: AstNode[];
+  parent?: AstNode<undefined>;
+}
+
+export interface ArrayNode<T> {
```
aomarks, line 32: Wondering if `AstNode` could accept `ValueTypes | ValueTypes[]` instead of there being a separate interface for `ArrayNode`? (I'm sure there's a reason, though).

src/util/ast.ts
```diff
@@ -0,0 +1,105 @@
...
+export {ParseError} from 'jsonc-parser';
+
+type ValueTypes = string | number | boolean | null | undefined;
+
+/**
+ * A JSON AST node.
+ *
+ * A safer override, preferring unknown over any.
+ */
+export interface AstNode<T extends ValueTypes = ValueTypes>
```
aomarks, line 25: Consider `JsonAstNode` or even `JsonNode`? I sort of feel that knowing this is connected to JSON is the more useful thing to know.

src/script.ts
```diff
@@ -30,7 +32,7 @@ export interface ScriptConfig extends ScriptReference {
    * An undefined command is valid as a way to give name to a group of other
    * scripts (specified as dependencies).
    */
-  command: string | undefined;
+  command: AstNode<string> | undefined;
```
aomarks, line 35: Threading the `AstNodes` through into the resolved script configs implies that it will be useful to have access to them in the `Executor`.

I think that _most_ of the errors for which it will be highly useful to have JSON squigglies would be caught in the `Analyzer`, since those are the errors that always require you to update the wireit config.

In the `Executor`, it seems maybe slightly less useful to show JSON squigglies, because most of the errors you encounter there are errors from the underlying script processes, rather than the build config, and usually the fix is not related to the wireit config. I could still seem some utility though, like if the error was because my command was wrong, this would let me quickly jump to it. Is that the kind of thing you were thinking?

aomarks (approved):
Great! Looking forward to seeing squigglies down the road.

aomarks (changes requested):
(Sorry, clicked wrong button -- I do have a few comments)

src/script.ts
rictic: My thinking is that, to provide editor features we'll want down the line, we'll want to keep the relevant ASTNodes around on the ScriptConfig objects. I agree that _probably_ the executor won't have much use for them, but the typings are good enough that I don't think it'll be a problem

aomarks: Sgtm

src/util/ast.ts
rictic: Sure. My thinking here was just that there's only one kind of AST that we're dealing with, but JsonAstNode is less ambiguous

src/util/ast.ts
rictic: Added a comment

src/script.ts
rictic: Yeah, that's reasonable. We don't need a `clean` AST node anyways. Done

src/script.ts
```diff
@@ -66,7 +66,7 @@ export interface ScriptConfig extends ScriptReference {
    * - "if-file-deleted": If an input file has been deleted, and before restoring from
    *   cache.
    */
-  clean: AstNode<boolean | 'if-file-deleted'> | undefined;
+  clean: boolean | 'if-file-deleted';
```
aomarks, line 69: Wondering if it would make sense to be an AstNode but excluding the undefined?

aomarks (approved):
LGTM! One question.

src/script.ts
rictic: So far I've avoided mutating or creating fake AstNodes. What would the range be? Would we make a full fake property with a fake key and value?

A problem we can chase if anything ever wants to refer to clean's location

PR #153, opened 2022-04-28
test: add some missing tests

43081j:
- Multiple cross-package dependencies
- Dependency chain in one package that fails in nested dependency
- Multiple errors occurring in usage
- Using an empty `WIREIT_PARALLEL` env variable

was just trying to get my head around how this all works, so here's some tests.

let me know if any of these don't make sense. the code they cover wasn't covered before.

### DIFF
--- a/src/test/basic.test.ts
+++ b/src/test/basic.test.ts
@@ -927,4 +927,114 @@ test(
   })
 );
 
+test(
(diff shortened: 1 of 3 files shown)


aomarks (approved):
This is great! Some really good situations to have coverage for in the tests. :tada:

PR #155, opened 2022-04-28
Prepare to release

rictic:
- adds a logo for the extension
- builds the extension in a subdirectory
- claims the extension id `google.wireit`
- updates some docs


vscode-extension/built/package.json
```diff
@@ -0,0 +1,34 @@
...
+        "fileMatch": "package.json",
+        "url": "./schema.json"
+      }
+    ]
+  },
+  "main": "./client.js",
+  "activationEvents": [
+    "onLanguage:json"
+  ],
+  "icon": "logo.png",
```
aomarks, line 20: (Can't comment on the png).

What happened to the squiggly part of the logo?

vscode-extension/src/scripts/copy-to-built.ts
```diff
@@ -0,0 +1,24 @@
...
+// npm workspaces, vsce, and npm starting at some version >8.12.2 and <=8.5.0
+// which causes issues with the current working directory when running scripts
+// in a workspace.
+// TODO(rictic): repro and file that bug
+
+try {
+  fs.mkdirSync('built');
+} catch {
+  // eslint-disable-next-line no-empty
+}
```
aomarks, line 20:
```suggestion
fs.mkdirSync('built', {recursive: true});
```

Setting `recursive: true` makes the already-exists case a non-error (https://nodejs.org/api/fs.html#fspromisesmkdirpath-options).

aomarks (changes requested):
Looks like the Windows test timed out. Probably need to bump to 30 mins.

Upping the parallelism from 1 will probably help here, but I want to implement stdio exclusivity locking first, because the log is really hard to read otherwise. (This is where we only let one script emit to stdout/stderr at a time, and buffer the other ones).

vscode-extension/built/package.json
rictic: I cropped it out here, because the main place it's displayed is in a small square in the extension sidebar, and just the text is already crowded there.

Sidebar with different color schemes:

Screen Shot 2022-04-28 at 11 12 41 AM
Screen Shot 2022-04-28 at 11 13 23 AM
Screen Shot 2022-04-28 at 11 14 01 AM

Full page (after clicking on the extension):
Screen Shot 2022-04-28 at 11 13 01 AM
Screen Shot 2022-04-28 at 11 13 30 AM
Screen Shot 2022-04-28 at 11 14 06 AM

vscode-extension/CHANGELOG.md
```diff
@@ -4,7 +4,9 @@ All notable changes to the "wireit" extension will be documented in this file.
 
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
-## [Unreleased]
+## v0.1.0 - 2022/04/28
```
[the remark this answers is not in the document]
rictic: Done

vscode-extension/built/package.json
aomarks: Ah ok good call. We could work on an even better suited logo sometime in the future maybe (just the W?). This seems great for now.

aomarks (approved):

Issue #154, opened 2022-04-28
Node 14.x Support (Vercel)


mihar-22:
I know it's wild but Vercel is still on `14.x`. Currently deploys fail because of newer syntax such as logical nullish assignment being included in the distribution.

![image](https://user-images.githubusercontent.com/14304599/165752004-958cb48a-4281-4982-8f18-e2b00d7ce137.png)

brendankenny:
FWIW we were just taking a look at wireit for our increasingly hard-to-manage npm script ordering over in [lighthouse](https://github.com/GoogleChrome/lighthouse). Our policy has settled over the years to supporting the Active and _one_ Maintenance LTS (back to Node 14, currently), so this is a blocker for us.

The tsconfig switch to `es2020` would be trivial, but it looks like there are a few Node 15+ features that might be annoying/not worth it to unwind (`AggregateError` and `str.replaceAll` especially). If you did want to expand support, maybe the easiest path would be to not remove Node 16 when it switches to maintenance in October and ease into supporting a maintenance release that way?

aomarks:
Very fair, happy to expand support to cover 14. It is still a supported LTS after all.

You identified all the issues, yeah. All petty trivial. Just need a ponyfill for AggregateError. Let's just write an inline one instead of adding a dep, since it's so simple.

aomarks:
Just published `0.3.0` which supports Node 14.14.0 and above.

brendankenny:
Awesome, thank you @aomarks!

PR #160, opened 2022-04-29
Support Node 14

aomarks:
Drops our minimum supported Node version from 16 to 14. Node 14 is the most recent [Node Maintenance LTS release](https://nodejs.org/en/about/releases/), with end-of-life on 2023-04-30.

Required the following changes:

- Language changes

  - Emit `es2020` instead of `es2022` from `tsc`.

  - Ponyfill `AggregateError`.

  - Replace uses of `AbortController` with `Deferred`. We could have pony/polyfilled `AbortController`, but the semantics feel so close to a `Deferred` that it didn't really seem worthwhile.

- npm changes

  - npm 6 prints more verbose details to `stderr` when a process fails, so I made our error-output checks less strict (`match` instead of `equals`).

  - npm 6 doesn't support workspaces, so we `skip` the workspaces test on Node 14.

  - npx 6 doesn't set `npm_` environment variables, so we can't directly detect `npx`. It still correctly errors, just with a different message.

  - npm 6 only includes the immediate package's `node_modules/.bin` directory in `$PATH`. npm 7+ also includes all of the parent `node_modules/.bin` directories. We rely on the newer behavior in a few tests, so the test rig now manually augments the `$PATH` to account for this difference.

Also hit a bootstrapping issues for the tests! Since we test Wireit itself with the latest version of Wireit from npm, but that version doesn't currently support Node 14, we can't use it to run this PR's tests. So I'm just manually invoking the test commands for now, and after this merges and is published, we can go back to self-hosting.

Fixes https://github.com/google/wireit/issues/154


package.json
```diff
@@ -216,7 +216,7 @@
     }
   },
   "devDependencies": {
-    "@types/node": "^17.0.23",
+    "@types/node": "^14.0.0",
```
AndrewJakubowicz, line 219: Unrelated to this change: Will dependabot try to bump the major version?

src/test/util/test-rig.ts
```diff
@@ -141,6 +142,26 @@ export class WireitTestRig extends FilesystemTestRig {
       WIREIT_CACHE: undefined,
       ACTIONS_CACHE_URL: undefined,
       ACTIONS_RUNTIME_TOKEN: undefined,
+      // In npm 6 (which ships with Node 14), "npm run" only includes the
+      // immediate package's "node_modules/.bin" directory in $PATH. Above npm
+      // 6, all of the parent ".bin" directories are included too.
```
AndrewJakubowicz, line 147: Does this also create additional considerations when using wireit with npm 6 with parent/child packages for external users?

AndrewJakubowicz (approved):
Really thorough and great PR description. Made it very clear why each change was made. Thank you!

src/test/util/test-rig.ts
aomarks: I don't think so, that's just how npm 6 worked for any installed binary. It would be the same if you tried to execute `tsc` or `eslint` or whatever from a child package. I think this npm change to `$PATH` was probably made as part of adding the workspaces feature.

package.json
aomarks: Good call!

I found that you can control which kinds of updates you take for specific packages here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore

Added config so that it won't suggest major updates to `@types/node`.

.github/dependabot.yml
```diff
@@ -8,6 +8,11 @@ updates:
     directory: '/'
     schedule:
       interval: 'daily'
+    ignore:
+      # Since we support older Node versions, we want our types to reflect the
+      # lowest supported versions of Node APIs.
+      - dependency-name: '@types/node'
+        update-types: ['version-update:semver-major']
```
AndrewJakubowicz, line 15: Awesome!

package.json
```diff
@@ -19,7 +19,7 @@
     "schema.json"
   ],
   "engines": {
-    "node": ">=16.7.0"
+    "node": ">=14.0.0"
```
augustjk, line 22: `fs.rm` was added in node 14.14 so I think that should be noted as the minimum. There were others like top level await in 14.8, and some import issues until 14.13, but I found 14.14 was lowest minor version that would pass the tests.

src/test/errors-analysis.test.ts
```diff
@@ -1180,8 +1180,8 @@ test(
     const result = rig.exec('npm run a');
     const done = await result.exit;
     assert.equal(done.code, 1);
-    assert.equal(
-      done.stderr.trim(),
+    assert.match(
```
rictic, line 1183: What changed to cause the swap from equal to match?

rictic (approved):

src/test/errors-analysis.test.ts
rictic: Ah, I see it in the PR description

src/util/aggregate-error.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Ponyfill for ES2022 AggregateError.
+ *
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError
+ * https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-aggregate-error-objects
+ */
+export class AggregateError {
+  errors: unknown[];
+  message?: string;
```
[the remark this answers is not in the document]
aomarks: Hm yeah, not needed. Done!

package.json
aomarks: Oh, great catch! Thank you for running the tests!

Done.

augustjk (approved):

CHANGELOG.md
```diff
@@ -6,7 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Changed
+
+- The minimum Node version is now `14.0.0` instead of `16.0.0`.
```
augustjk, line 13:
```suggestion
- The minimum Node version is now `14.14.0` instead of `16.0.0`.
```

augustjk: oops, i was a little late on this. maybe it can be fixed when making a PR for the release.

aomarks: Oops, sorry I missed this. Fixed in https://github.com/google/wireit/pull/161.

Icelamar (commented):
[]()

Icelamar (commented):
[@]()

PR #162, opened 2022-04-29
Support Node 14 on Windows and release 0.3.1

aomarks:
I didn't test Node 14 on Windows in https://github.com/google/wireit/pull/160, so I missed two issues:

- We use `replaceAll`, but only on Windows. Fixed by using `replace` instead, and by dropping `lib` down to `es2020` in `tsconfig.json` (which catches this error in `tsc`).

- There is a subtle difference between how environment variables are treated when `spawn`ing between Node 14 and 16 on Windows, relating to the way environment variables are case-insensitive on Windows. Full details in `test/windows.ts`.

I've also expanded to the test matrix to just test all OS/Node permutations that we support, because obviously it is not safe to assume anything (especially on Windows, but let's be safe and not assume anything about macOS acting exactly like Linux either).

Actually fixes https://github.com/google/wireit/pull/160

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,7 @@
     "module": "es2022",
     "moduleResolution": "node",
     "useDefineForClassFields": true,
-    "lib": ["es2021"],
+    "lib": ["es2020"],
     "rootDir": "src",
     "outDir": "lib",
     "strict": true,
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,29 +5,11 @@ on: [push, pull_request]
 jobs:
   tests:
     strategy:
+      # We support Node Latest, Active LTS, and the most recent Maintenance LTS.
+      # See https://nodejs.org/en/about/releases/ for Node release schedule.
       matrix:
-        # See https://nodejs.org/en/about/releases/ for Node release
-        # information.
-        include:
-          # Most recent Node Maintenance LTS
-          - os: ubuntu-20.04
-            node: 14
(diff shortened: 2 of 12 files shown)


augustjk (approved):

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "module": "es2022",
```
AndrewJakubowicz, line 4: Does this also need to be addressed for node 14? es2022 allows top level await, which was added in [Node v14.8.0](https://nodejs.org/en/blog/release/v14.8.0/#:~:text=Derek%20Lewis)%20%2334602-,%5B62bb2e757f%5D,-%2D%20(SEMVER%2DMINOR)%20module)

AndrewJakubowicz (approved):

tsconfig.json
aomarks: Our min version is currently actually `14.14.0`, so we can use top-level await.

I'm hoping people who use npm 14 at least use a _recent_ npm 14 -- if they don't then we'll have to also drop top-level await, `fs.rm`, and anything else that wasn't in `node 14.0.0`.

aomarks:
I discovered that there was an additional extremely gnarly problem relating to the way environment variable case sensitivity works in Windows. There was a subtle difference between Node 14 and 16, which is now fixed!

PTAL, non trivial new code!

augustjk (approved):
Good find! That's some interesting behavior with the environment variables.

aomarks:
Thank you! Published.

PR #165, opened 2022-05-01
Add and test a function to draw squiggles under text

rictic:
This is tricky enough code that it's worth checking in separate.

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -27,3 +27,39 @@ export class WireitError extends Error {
     this.event = event;
   }
 }
+
+export interface Range {
+  readonly offset: number;
+  readonly length: number;
+}
+
+// Exported for testing
+export function drawSquiggleUnderRange(
+  range: Range,
+  fileContents: string,
+  indent: number
+): string {
+  let {offset, length} = range;
+  const startOfInitialLine =
+    fileContents.slice(0, offset).lastIndexOf('\n') + 1;
+  const uncorrectedFirstNewlineIndexAfter = fileContents
+    .slice(offset + length)
+    .indexOf('\n');
+  const endOfLastLine =
+    uncorrectedFirstNewlineIndexAfter === -1
+      ? undefined
+      : offset + length + uncorrectedFirstNewlineIndexAfter;
+  offset = offset - startOfInitialLine;
+
+  const sectionToPrint = fileContents.slice(startOfInitialLine, endOfLastLine);
+  let result = '';
+  for (const line of sectionToPrint.split('\n')) {
+    result += `${' '.repeat(indent)}${line}\n`;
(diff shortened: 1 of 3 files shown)


src/test/diagnostic.test.ts
```diff
@@ -0,0 +1,143 @@
...
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {suite} from 'uvu';
+import * as assert from 'uvu/assert';
+import {drawSquiggleUnderRange} from '../error.js';
+
+const test = suite();
```
aomarks, line 11:
```suggestion
import {test} from 'uvu';
import * as assert from 'uvu/assert';
import {drawSquiggleUnderRange} from '../error.js';
```

src/test/diagnostic.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  assert.equal(
+    drawSquiggleUnderRange({offset: 3, length: 3}, 'aaabbbccc\nabc', 8),
+    `
+        aaabbbccc
+           ~~~`.slice(1)
+  );
+});
+
+test('drawing squiggles under multi-line ranges', () => {
```
aomarks, line 85: This has the same test name as the one above

src/error.ts
```diff
@@ -27,3 +27,39 @@ export class WireitError extends Error {
...
+      ? undefined
+      : offset + length + uncorrectedFirstNewlineIndexAfter;
+  offset = offset - startOfInitialLine;
+
+  const sectionToPrint = fileContents.slice(startOfInitialLine, endOfLastLine);
+  let result = '';
+  for (const line of sectionToPrint.split('\n')) {
+    result += `${' '.repeat(indent)}${line}\n`;
+    const squiggleLength = Math.min(line.length - offset, length);
+    result += ' '.repeat(offset + indent) + '~'.repeat(squiggleLength) + '\n';
```
aomarks, line 59: Can be for later, but would be cool to print the squiggles in red. I think it could be as simple as something like:

```ts
const RED = '\x1b[31m';
const RESET = '\x1b[0m';
...
result += ' '.repeat(offset + indent) + RED + '~'.repeat(squiggleLength) + RESET + \n';
```

aomarks (approved):
Nice!

src/test/diagnostic.test.ts
rictic: Ah, fixed

src/test/diagnostic.test.ts
rictic: Done

src/error.ts
rictic: +1, will do in a follow up PR

PR #167, opened 2022-05-02
Upgrade tests to use pnpm7

aomarks:
Supersedes https://github.com/google/wireit/pull/166

Our pnpm workspaces test was affected by this change in pnpm 7:

```
When using pnpm run <script>, all command line arguments after the script name are now passed
to the script's argv, even --. For example, pnpm run echo --hello -- world will now pass
--hello -- world to the echo script's argv. Previously flagged arguments (e.g. --silent)
were interpreted as pnpm arguments unless -- came before it.
```

https://github.com/pnpm/pnpm/releases/tag/v7.0.0

Also fixed an incorrect date in the CHANGELOG

### DIFF
--- a/src/test/basic.test.ts
+++ b/src/test/basic.test.ts
@@ -910,7 +910,7 @@ test(
 
     // Run both from the workspaces root package.
     {
-      const exec = rig.exec('pnpm run cmd --recursive');
+      const exec = rig.exec('pnpm run --recursive cmd');
       // Workspace commands run in serial.
       (await cmdA.nextInvocation()).exit(0);
       (await cmdB.nextInvocation()).exit(0);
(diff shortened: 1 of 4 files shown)


src/test/basic.test.ts
```diff
@@ -910,7 +910,7 @@ test(
 
     // Run both from the workspaces root package.
     {
-      const exec = rig.exec('pnpm run cmd --recursive');
```
AndrewJakubowicz, line 913: Re-iterating what you wrote in the description. This broke because `--recursive` ended up passed in as an argv as it came after the script name in pnpm 7. Re-ordering removes `--recursive` from the commands argv.

augustjk (approved):

AndrewJakubowicz (approved):
Nice fix!

src/test/basic.test.ts
aomarks: Yep!

PR #171, opened 2022-05-03
Add an offset to position converter

rictic:
Parsers, including jsonc-parser, often give locations in terms of a character offset, but when communicating to users we want to use line and character numbers.

This also introduces a JsonFile object, which includes the path, the string contents, and the ast of a json file all in one object. Useful for passing around, and also, in an upcoming change, as a WeakMap key to cache the OffsetToPosition converter from.

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -5,6 +5,7 @@
  */
 
 import type {Failure} from './event.js';
+import {JsonFile} from './util/package-json-reader.js';
 
 /**
  * A known Wireit error.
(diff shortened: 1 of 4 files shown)


src/error.ts
```diff
@@ -33,13 +34,55 @@ export interface Range {
...
+
+export interface Position {
+  /** 1 indexed */
+  line: number;
+  /** 1 indexed */
+  column: number;
+}
+
+export class OffsetToPositionConverter {
+  readonly newlineIndexes: readonly number[];
```
aomarks, line 50: `#private`?

aomarks (approved):

src/error.ts
rictic: This is useful for other stuff too, e.g. could potentially make drawSquiggle more efficient

PR #176, opened 2022-05-04
Add some color to our diagnostics.

rictic:


src/test/util/colors.ts
```diff
@@ -0,0 +1,5 @@
+
```
aomarks: License header missing

aomarks (approved):
Looks fantastic!

image

src/test/util/colors.ts
rictic: Eep! Done

PR #177, opened 2022-05-04
Convert invalid config error to diagnostic

rictic:
This is one of the most common error types, so the change is pretty big, but hopefully legible.


aomarks (approved):
Nice!

Issue #173, opened 2022-05-03
Globs should be forced relative to package directory


aomarks:
Currently `/foo` means exactly that. The file called "foo" in the root of the filesystem. This is different to how the npm `files` array works, which "re-roots" absolute paths on-top of the current package (in other words, it ignores the leading slash). We should work the same way, since it's currently a big footgun.

One special case is that for input files, we do want to support selecting files outside of the current package. So we should still support `../`.

Output files can't currently support `../` because of the way we copy output to the `.wireit/<script>/cache` folder. We could support it by representing local cache differently (e.g. with a tarball).

PR #180, opened 2022-05-04
Interpret leading slashes relative to package

aomarks:
Previously, a glob pattern like `/foo` was resolved to `foo` in the root of the filesystem, which is the usual behavior for a file path.

However, that's a footgun because what a user almost certainly means by that is `foo` in the package directory.

So we now ignore the leading `/`. In the case of `files`, we still allow reaching out of the directory using e.g. `../foo`.

This new behavior also matches the behavior of the npm `files` array (apart from the `../` support, which `files` simply ignores).

Also fixed a duplicate `lint` script in `package.json`.

Fixes https://github.com/google/wireit/issues/173

### DIFF
--- a/src/test/glob.test.ts
+++ b/src/test/glob.test.ts
@@ -508,11 +508,12 @@ test('re-rooting allows ../', ({check}) =>
     expected: ['../foo'],
   }));
 
-test('re-rooting handles /./foo', ({check}) =>
+// TODO(aomarks) This should be normalized to "foo" consistently. It currently
+// differs on Windows between Node 14 and 16.
+test.skip('re-rooting handles /./foo', ({check}) =>
   check({
     files: ['foo'],
     patterns: ['/./foo'],
-    // TODO(aomarks) This should be normalized to to "foo".
     expected: [`.${pathlib.sep}foo`],
   }));
 
--- a/src/test/glob.test.ts
+++ b/src/test/glob.test.ts
@@ -28,6 +28,7 @@ interface TestCase {
   followSymlinks?: boolean;
   includeDirectories?: boolean;
   expandDirectories?: boolean;
+  rerootToCwd?: boolean;
 }
 
 const test = suite<{
@@ -44,11 +45,12 @@ test.before.each(async (ctx) => {
       files,
       patterns,
       expected,
-      cwd = rig.temp,
+      cwd = '.',
       absolute = false,
       followSymlinks = true,
       includeDirectories = false,
       expandDirectories = false,
+      rerootToCwd = false,
     }: TestCase): Promise<void> => {
       for (const file of files) {
         if (typeof file === 'string') {
@@ -73,11 +75,12 @@ test.before.each(async (ctx) => {
       let actual, error;
       try {
         actual = await glob(patterns, {
-          cwd,
+          cwd: rig.resolve(cwd),
           absolute,
           followSymlinks,
           includeDirectories,
           expandDirectories,
+          rerootToCwd,
         });
       } catch (e) {
         error = e;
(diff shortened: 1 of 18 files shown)


src/util/glob.ts
```diff
@@ -131,8 +144,14 @@ export async function glob(
...
   // way to do that is to build the groups backwards.
   for (let i = expandedPatterns.length - 1; i >= 0; i--) {
-    const pattern = expandedPatterns[i];
+    let pattern = expandedPatterns[i];
     const isExclusive = pattern[0] === '!';
+    if (isExclusive) {
+      pattern = pattern.slice(1); // Remove the "!"
+    }
+    if (opts.rerootToCwd) {
+      pattern = pattern.replace(/^\/+/, '');
```
augustjk, line 153: Is the `+` necessary in the regex here? I suppose it doesn't hurt to match extra if the user happens to put in something like `//foo`.

augustjk (approved):

src/util/glob.ts
aomarks: Yeah, if the user did that (which is a weird but valid path) we actually _must_ remove both `//`s, or else we'd still interpret it relative to the filesystem root.

augustjk: Ah, okay did not know multiple slashes would be considered valid. TIL it's just treated like a single slash.

src/util/glob.ts
```diff
@@ -131,8 +144,14 @@ export async function glob(
   // We want each group to include all subsequent negated patterns. The simplest
   // way to do that is to build the groups backwards.
   for (let i = expandedPatterns.length - 1; i >= 0; i--) {
-    const pattern = expandedPatterns[i];
+    let pattern = expandedPatterns[i];
     const isExclusive = pattern[0] === '!';
+    if (isExclusive) {
+      pattern = pattern.slice(1); // Remove the "!"
+    }
+    if (opts.rerootToCwd) {
```
AndrewJakubowicz, line 152: Should this be an option or enabled by default? All use cases (except tests) pass in `rerootToCwd = == true`.

aomarks: True. I've just made it the only behavior, since we want it everywhere.

PTAL

src/util/glob.ts
aomarks: Added a comment to clarify this.

src/test/glob.test.ts
```diff
@@ -486,4 +486,40 @@ test('dirent tags symlinks to directories as directories when followSymlinks=tru
...
+    expected: ['foo', 'bar'],
+  }));
+
+test('braces can be escaped', ({check}) =>
+  check({
+    files: ['{foo,bar}'],
+    patterns: ['\\{foo,bar\\}'],
+    expected: ['{foo,bar}'],
+  }));
+
```
rictic, line 524: No one should write this, but how do we interpret /./foo or /../bar? Probably worth a test to make sure what we do is sensible

rictic (approved):
One request

src/test/glob.test.ts
aomarks: We remove the leading `/`, which I think is correct.

Added tests.

AndrewJakubowicz (approved):
Nice!

AndrewJakubowicz:
Failing Windows test - on Node 14

aomarks:
> Failing Windows test - on Node 14

Fixed

rictic (approved):

aomarks:
> > Failing Windows test - on Node 14
> 
> Fixed

Actually not fixed, it's inconsistent on Windows between 14 and 16. But it's an inconsistency between `foo` and `./foo` which are semantically the same, and my next PR (https://github.com/google/wireit/pull/182) just so happens to fix this by adding extra normalization. Temporarily disabled this one test.

Issue #64, opened 2022-04-01
Detect output outside of package dir during analysis instead of execution


aomarks:
As a safety precaution, we refuse to clean output files if they are outside of the current script's package. However, it would be even better to catch this error earlier, in the Analyzer, by looking at the `output` glob patterns. We just need to be careful that we are parsing and analyzing the glob patterns correctly, accounting for negations, `{}` groups, and other special glob syntax.

PR #182, opened 2022-05-04
Error when trying to cache outside of package

aomarks:
It's currently not possible to locally cache an output file that isn't inside of the package directory. We check for this case when we delete and throw, but not when we cache. So if you are caching but have cleaning disabled, we would silently weirdly save the output file to a parent directory, and then not restore it.

Now this is an error.

Note we could in theory do this during analysis, but I'm not 100% confident in my ability to correctly detect this case given all of the possible magic glob syntax, so for now it's safer to just do it at runtime. (see https://github.com/google/wireit/issues/64).

Also note we could in theory support caching files outside of the package root, but we'd have to do something like a tarball for the local cache, instead of simply copying into `.wireit/<script>/cache/<hash>`. We should think carefully about whether we want to do that, though, so I'm not dealing with that for now.

Fixes https://github.com/google/wireit/issues/181

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -492,6 +492,7 @@ class ScriptExecution {
           followSymlinks: false,
           includeDirectories: true,
           expandDirectories: true,
+          rerootToCwd: true,
         }
       )
     );
@@ -530,6 +531,7 @@ class ScriptExecution {
         // We must expand directories here, because we need the complete
         // explicit list of files to hash.
         expandDirectories: true,
+        rerootToCwd: true,
       });
       // TODO(aomarks) Instead of reading and hashing every input file on every
       // build, use inode/mtime/ctime/size metadata (which is much faster to
@@ -662,6 +664,7 @@ class ScriptExecution {
       followSymlinks: false,
       includeDirectories: true,
       expandDirectories: true,
+      rerootToCwd: true,
     });
     if (absFiles.length === 0) {
       return;
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -15,7 +15,7 @@ import {shuffle} from './util/shuffle.js';
 import {WorkerPool} from './util/worker-pool.js';
 import {getScriptDataDir} from './util/script-data-dir.js';
 import {unreachable} from './util/unreachable.js';
-import {glob} from './util/glob.js';
+import {glob, GlobOutsideCwdError} from './util/glob.js';
 import {deleteEntries} from './util/delete.js';
 import {AggregateError} from './util/aggregate-error.js';
 import {
@@ -458,10 +458,9 @@ class ScriptExecution {
     if (this.#cache === undefined || this.#script.output === undefined) {
       return;
     }
-    await this.#cache.set(
-      this.#script,
-      stateStr,
-      await glob(
+    let paths;
+    try {
+      paths = await glob(
         [
           ...this.#script.output.values,
           // Also include the "stdout" and "stderr" replay files at their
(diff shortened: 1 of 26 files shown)


src/util/glob.ts
```diff
@@ -190,14 +205,20 @@ export async function glob(
...
       const matches = await fastGlob(include, {
         ignore: exclude,
-        cwd: opts.cwd,
+        cwd: normalizedCwd,
         dot: true,
         onlyFiles: !opts.includeDirectories,
-        absolute: opts.absolute,
+        // Return absolute paths, even if we ultimately return relative ones, so
+        // that we can do path string comparisons.
+        absolute: true,
```
augustjk, line 221: Is the path string comparison only necessary if `opts.throwIfOutsideCwd` is true? I wonder if it would be much more complex to only override the `absolute` passed to `fastGlob` and revert to relative paths if needed, only if `throwIfOutsideCwd` is true.

src/util/glob.ts
```diff
@@ -212,20 +233,44 @@ export async function glob(
...
+        //
+        // 2. Trailing slashes and other remnants of the input patterns like
+        //    ".." are normalized away. fast-glob tends to preserve these in the
+        //    results. Note that `fs.resolve` trims trailing slashes, but
+        //    `fs.normalize` does not.
+        let path = pathlib.resolve(match.path);
+        if (
+          opts.throwIfOutsideCwd &&
+          !path.startsWith(normalizedCwdWithTrailingSep) &&
+          path !== normalizedCwd
```
augustjk, line 249: There's probably something I'm not thinking of but why do we need these 2 separate checks instead of just `!path.startsWith(normalizedCwd)`?

src/executor.ts
```diff
@@ -658,33 +674,34 @@ class ScriptExecution {
...
+        cwd: this.#script.packageDir,
+        absolute: true,
+        followSymlinks: false,
+        includeDirectories: true,
+        expandDirectories: true,
+        rerootToCwd: true,
+        throwIfOutsideCwd: true,
+      });
+    } catch (error) {
+      if (error instanceof GlobOutsideCwdError) {
```
rictic, line 689: I think, for tools like this that are often misconfigured and where we want to give a clear error message in that case, throwing errors is probably a bad strategy for communicating known failure modes, because it's not particularly type safe or easy to remember / notice.

WDYT about starting to use Result types instead?

rictic:
I think after rebasing, the new errors will also need to add diagnostics

rictic (approved):

src/executor.ts
aomarks: Might be right, but I think I'd prefer to think about a refactoring like this some other time.

I decided to use use exceptions for errors consistently because it seemed like it simplified error handling. Exceptions can always happen because of APIs we don't control, so they are in the mix regardless. So we use `AggregateError` to accumulate all of the errors we encounter as we execute the build graph (both known and unknown). And then at the top-level, we can just look at all the errors, and nicely format the known ones. I'm not sure that's so obviously bad.

src/util/glob.ts
```diff
@@ -208,20 +227,40 @@ export async function glob(
...
+        // 2. Remnants of input pattern syntax like ".." and trailing "/"s are
+        //    removed (which fast-glob preserves in the results). Note that
+        //    `fs.normalize` does not trim trailing "/"s, so we do that
+        //    ourselves (`fs.resolve` does, but that also makes the path
+        //    absolute).
+        match.path = pathlib.normalize(match.path.replace(/\/+$/g, ''));
+        if (opts.throwIfOutsideCwd) {
+          const absPath = opts.absolute
+            ? match.path
+            : pathlib.resolve(normalizedCwd, match.path);
```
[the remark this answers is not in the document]
aomarks: Yeah, `absolute:true` seems to remove the weird remnants. But it's possible there are cases where it doesn't, so I feel better normalizing always.

PR #185, opened 2022-05-04
Release v0.2.0 of the extension.

rictic:


vscode-extension/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to the "wireit" extension will be documented in this file.
 
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
-## Unreleased
+## [0.2.0] - 2022-05-04
```
aomarks, line 7: I think maybe should be `0.1.1` because it's non-breaking and the leftmost is 0?

aomarks (approved):

vscode-extension/CHANGELOG.md
rictic: VSCode extension version numbers aren't semver because nothing else uses them as a library. IMO they're marketing, and this feels like a 0.2 to me as it adds major new functionality

Issue #33, opened 2022-03-22
Service mode


aomarks:
By default, Wireit assumes that scripts eventually exit by themselves. This works well for things like building or testing. But sometimes a script runs indefinitely, such as a server.

Setting `"server": true` will tell Wireit that a script runs indefinitely. This has the following effects:

- It will always run. It will never be skipped or restored from cache.

- If something depends on the server, Wireit won't wait for the server to exit before the dependent script starts running. Instead, it just waits for the server process to be spawned.

- If a server script is run _directly_ (e.g. `npm run serve`), then it will stay running until the user kills Wireit (e.g. `Ctrl-C`).

- If a server script is run _indirectly_ (e.g. `npm run script-that-depends-on-server`), then the server script will stay running until all scripts which transitively depend on it have finished.

- In watch mode, Wireit will restart the server whenever a dependency changes. If this isn't required for a particular dependency (such as for static assets that the server does not cache), the dependency edge can be annotated with `"restart": false`.

```json
{
  "scripts": {
    "build": "wireit"
  },
  "wireit": {
    "serve": {
      "command": "node lib/server.js",
      "server": true,
      "dependencies": [
        "build:server",
        {
          "script": "build:assets",
          "restart": false
        }
      ],
      "files": [],
      "output": []
    }
  }
}
```

justinfagnani:
I did eventually run into a spot where I needed to start a server to run an integration test...

aomarks:
> I was trying to set up wireit to work for running a server in watch mode for development, and the issue I ran into is that the watch mode doesn't start watching again until after the server is shut down, which doesn't work well for a dev environment.

Yeah, this is a high priority issue.

In your case, would restarting your server on each watch iteration work?

I was also thinking that you often have some dependencies which require a restart, and some which don't. So maybe you declare like this:

```json
"wireit": {
  "my-server": {
    "server": true,
    "dependencies": [
      "build-server-code",
      {
        "name": "server-assets",
        "restart-on-change": false
      }
    ]
  }
}
```

> I also really love the if-file-deleted feature as that solves a huge issue I've been frustrated with.

🙂

ZachHaber:
I feel like restart-on-change would be problematic and redundant with "files". 

I.e. `"files": []` would indicate that it shouldn't ever restart. `"files": ["*"]` would presumably indicate that it'd restart when anything changes.

[...]

PR #186, opened 2022-05-04
Handle keyboard input in child process

webpro:
When wiring release-it up with wireit and create a monorepo setup, I saw interesting use cases.

However, using a process like this would require a way to set `stdio: 'inherit'`, or deal in some other way with keyboard input. Would you be open to something like this? Perhaps as a configuration option of sorts. Maybe this should even work out-of-the-box, although currently I'm not sure how handling this stream fits into the rest of the program.

Just to be sure, this first commit is not to be merged as-is, it's just to open this discussion in a draft PR.

Please see the readme and `package.json` of https://github.com/webpro/release-it-monorepo. Also see this screenshot for the working demo to publish three packages in this mini monorepo:

Screen Shot 2022-05-04 at 8 47 33 PM

By the way, release-it can be used in `--ci` mode as well, eliminating the need for user/keyboard input.

### DIFF
--- a/src/script-child-process.ts
+++ b/src/script-child-process.ts
@@ -13,7 +13,11 @@ import {
 
 import type {Result} from './error.js';
 import type {ScriptConfig} from './script.js';
-import type {ChildProcessWithoutNullStreams} from 'child_process';
+import type {
+  ChildProcessWithoutNullStreams,
+  ChildProcessByStdio,
+} from 'child_process';
+import {Readable} from 'stream';
 import type {ExitNonZero, ExitSignal, SpawnError, Killed} from './event.js';
 
 /**
@@ -56,7 +60,9 @@ type ScriptConfigWithRequiredCommand = ScriptConfig & {
  */
 export class ScriptChildProcess {
   readonly #script: ScriptConfigWithRequiredCommand;
-  readonly #child: ChildProcessWithoutNullStreams;
+  readonly #child:
+    | ChildProcessWithoutNullStreams
+    | ChildProcessByStdio<null, Readable, Readable>;
   #state: ScriptChildProcessState = 'starting';
 
   /**
@@ -90,6 +96,7 @@ export class ScriptChildProcess {
       //   https://nodejs.org/api/child_process.html#default-windows-shell
       //   https://github.com/npm/run-script/blob/a5b03bdfc3a499bf7587d7414d5ea712888bfe93/lib/make-spawn-args.js#L11
       shell: true,
+      stdio: ['inherit', 'pipe', 'pipe'],
       env: augmentProcessEnvSafelyIfOnWindows({
         PATH: this.#pathEnvironmentVariable,
       }),


aomarks:
Thanks for the PR! Definitely open to this or something along these lines, it's something I've been thinking we should support.

It seems like something that should be a setting. Perhaps like this?:

```json
{
  "wireit": {
    "publish": "release-it",
    "interactive": true
  }
}
```

There are some details to think about:

- What happens if two interactive processes try to run at the same time? Do we need to prevent this?
- What happens if something depends on an interactive process?
- What happens in watch mode? Seems also related to https://github.com/google/wireit/issues/33.

webpro:
> Thanks for the PR! Definitely open to this or something along these lines, it's something I've been thinking we should support.

OK great!

> It seems like something that should be a setting. Perhaps like this?:
> 
> ```json
> {
>   "wireit": {
>     "publish": "release-it",
>     "interactive": true
>   }
> }
> ```
> 
> There are some details to think about:
> 
> * What happens if two interactive processes try to run at the same time? Do we need to prevent this?

Yes, at first thought it makes sense to me to basically enforce `WIREIT_PARALLEL=1` for `interactive: true`. _Maybe_ this is basically the same thing, if `inherit` would be the default for `stdin`.

> * What happens if something depends on an interactive process?

I guess that should be automatically solved as well by enforcing a single/sequential flow in `interactive` mode.

> * What happens in watch mode?

Uh, interesting. No sure, tbh. Since `watch` is opt-in it's not an issue to be solved by wireit per se.

> * Seems also related to [Server mode #33](https://github.com/google/wireit/issues/33).

Using `inherit` for `stdio` might help there as well. I'll update this PR now with a better fix to use `inherit` only for the `stdin` stream. This should also make the tests pass.

webpro:
I just force-pushed the same, would love to make progress here

aomarks:
Note that as a workaround, which I'm sure you've already discovered, you can probably just chain the build + release scripts in the regular npm script. There may not be any advantage to having wireit execute the final command -- right?

```json
{
  "scripts": {
    "publish": "npm run build && release-it",
    "build": "wireit"
  }
}
```

webpro:
> > I just force-pushed the same (after the refactor between 0.3.1 and 0.4.1), would love to make progress here
> 
> Sorry for the delay, I still need to think carefully through the implications of this and exactly how it should work, to answer some of the questions above, such as ensuring that only one interactive script at a time receives `stdin`. I'm currently working on #33 so we'll need to think about interaction there too.

Alright, that's cool of course. No hurries, I'm just excited for this feature :)

> Note that as a workaround, which I'm sure you've already discovered, you can probably just chain the build + release scripts in the regular npm script. There may not be any advantage to having wireit execute the final command -- right?
> 
> ```json
> {
>   "scripts": {
>     "publish": "npm run build && release-it",
>     "build": "wireit"
>   }
> }
> ```

Yes, I'm aware of this. I'm especially interested in this specific feature, so users of [release-it](https://github.com/release-it/release-it) can use it for monorepos too. I think adding (something like) the changes in this PR can result in a great combination of both tools:

```json
{
  "scripts": {
    "release": "wireit"
  },
  "wireit": {
    "release": {
      "command": "release-it --no-npm",
      "dependencies": [
        "./packages/bravo:release",
        "./packages/mol:release",
        "./packages/zen:release"
      ]
    }
  }
}
```

[This monorepo](https://github.com/webpro/release-it-monorepo) serves as a demo. Without wireit, this does the same:

```json
{
  "scripts": {
    "release": "npm --prefix packages/bravo run release && npm --prefix packages/bravo run release && npm --prefix packages/zen run release && release-it --no-npm"
  }
}
```

Another option would be to build something like this into release-it itself, but that would add complexity that I'd rather prevent. I haven't found great alternatives myself so far, but when wireit ends up without supporting interactive mode I'll definitely consider that.

I'd also like to add that roughly 50% of release-it runs happen in CI mode, so without interactivity, which works like a charm today (even without `WIREIT_PARALLEL=1`, although output is obviously mixed up as a consequence).

aomarks:
Thank you for this PR!

I'm still planning to support interactive scripts at some point, but I think there's more design work to do. It can't be as simple as forwarding all stdin (as this PR does), because if there are multiple scripts running at the same time that take stdin, it won't be clear which one is "focused".

I expect we'll need to add an `"interactive": true` setting, which will ensure that only one interactive script is active at a time.

Tracking at https://github.com/google/wireit/issues/281

webpro:
Yes, that's totally clear and fair. Thanks!

Issue #48, opened 2022-03-28
Recover from analysis errors in watch mode


aomarks:
An analysis error encountered in watch mode doesn't need to terminate the wireit command. We can keep watching all known `package.json` files and keep iterating.

PR #188, opened 2022-05-05
Don't exit from analysis errors in watch mode 

rictic:


aomarks (changes requested):
It would be nice if all of the refactoring included here was in a separate PR.

rictic:
That's a great point, split out the errors to results part (the original point of this change) into its own PR and added this one on top to make watch mode resilient against analysis errors

src/watcher.ts
```diff
@@ -151,7 +151,14 @@ export class Watcher {
       for (const failure of analysisResult.error) {
         this.#logger.log(failure);
       }
-      process.exit(1);
+      // TODO: it would be good to get all the package.json files that
```
aomarks, line 154: There's a TODO right above here "Add support for recovering from..." that this comment supersedes.

src/watcher.ts
```diff
@@ -151,7 +151,14 @@ export class Watcher {
...
+      //   analyze read, so that we could watch them here.
+      //   I think we want to ultimately have analyze return as complete a
```
aomarks, line 156: I'm kinda thinking the "watch all the package.json files we encountered" aspect of this is important to really fix https://github.com/google/wireit/issues/48.

The effect now if you're working across packages, is that if you have an error in a different package, it's actually an extra step vs before, because now the watch mode just hangs after you fix the problem, so you have to CTRL-C before re-running. Before you still had to re-run, but I think it was more obvious that you'd need to do that, since the process exited.

WYDT about trying to handle that now? Makes sense to me for the `Analyzer` to return the set of all `package.json` files it encountered.

src/test/util/uvu-timeout.ts
```diff
@@ -6,7 +6,8 @@
 
 import type * as uvu from 'uvu';
 
-const DEFAULT_TIMEOUT = Number(process.env.TEST_TIMEOUT ?? 30_000);
+const TIMEOUT_IF_UNSET = process.env.CI ? 30_000 : 10_000;
```
aomarks, line 9: +1 I've been wanting this.

I'd actually maybe go even lower for local. I usually use 3000 which seems to work just fine on my computer. Maybe 5000 to be safe?

aomarks (changes requested):

src/watcher.ts
rictic: +1, I'll handle that in this PR, but I think I'm gonna want an abstraction I just developed in a different branch (a PackageJson class), and maybe some more work on not stopping on the first diagnostic we find, so I'll land this after that.

aomarks:
Superseded by https://github.com/google/wireit/pull/266

PR #189, opened 2022-05-06
Serialize execution of the same script across different Wireit processes

aomarks:
Previously, if you `npm run` multiple versions of the same script (or 2 Wireit scripts that share a dependency) at the same time, then they would simply run simultaneously, and incorrect output could be produced. In particular, the second Wireit process might try and delete output, while the first one is still writing it.

Now, we prevent multiple versions of the same script from running at the same time. Before any script starts, we try to acquire an exclusive lock for it, and we wait if the lock can't be acquired, logging this message:

```
💤  [a] Waiting for another process which is already running this script.
```

An exception is when `output=[]`, because I think these cases are unlikely to cause problems, and it allows users to continue running e.g. multiple versions of the same server or other non-build step if they need to.

The lockfile is managed using the [proper-lockfile](https://github.com/moxystudio/node-proper-lockfile) library, which seems to have a nice design. It works by attempting to `mkdir` a directory, which will either succeed if it doesn't exist, or fail it does exist, atomically. It also detects when a lock has probably been abandoned, by regularly updating the directory `mtime` when the lock is held, and deleting the lock directory if it hasn't been updated in a long time.

Fixes https://github.com/google/wireit/issues/35

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,12 @@ export WIREIT_PARALLEL=1
 npm run build
 ```
 
+If two or more seperate `npm run` commands are run for the same Wireit script
+simultaneously, then only one version will be allowed to run at a time, while
+the others wait their turn. This prevents coordination problems that can result
+in incorrect output files being produced. If `output` is set to an empty array,
+then this restriction is removed.
+
 ## Input and output files
 
 The `files` and `output` properties of `wireit.<script>` tell Wireit what your
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ npm run build
 ```
 
 If two or more seperate `npm run` commands are run for the same Wireit script
-simultaneously, then only one version will be allowed to run at a time, while
+simultaneously, then only one instance will be allowed to run at a time, while
 the others wait their turn. This prevents coordination problems that can result
 in incorrect output files being produced. If `output` is set to an empty array,
 then this restriction is removed.
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -22,6 +22,7 @@ import {
   augmentProcessEnvSafelyIfOnWindows,
   posixifyPathIfOnWindows,
 } from './util/windows.js';
+import lockfile from 'proper-lockfile';
 
 import type {
   ScriptConfig,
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -198,6 +198,12 @@ export WIREIT_PARALLEL=1
 npm run build
 ```
 
+If two or more seperate `npm run` commands are run for the same Wireit script
+simultaneously, then only one version will be allowed to run at a time, while
```
justinfagnani, line 202:
```suggestion
simultaneously, then only one instance will be allowed to run at a time, while
```
?

Talking about "versions" around npm could potentially be confused with package versions.

src/test/parallelism.test.ts
```diff
@@ -225,4 +225,80 @@ test(
...
+
+    const exec1 = rig.exec('npm run a');
+    const exec2 = rig.exec('npm run a');
+    const inv1 = await cmdA.nextInvocation();
+    const inv2 = await cmdA.nextInvocation();
+    inv1.exit(0);
+    inv2.exit(0);
+    assert.equal((await exec1.exit).code, 0);
+    assert.equal((await exec2.exit).code, 0);
+    assert.equal(cmdA.numInvocations, 2);
```
augustjk, line 300: I'm a little lost on how this tests that scripts are running without exclusive lock. Wouldn't this pass even if they ran one after another? Would asserting `numInvocations` to `2` before exiting be appropriate?

src/executor.ts
```diff
@@ -135,6 +136,79 @@ class ScriptExecution {
...
+            // Only log this once.
+            this.#logger.log({
+              script: this.#script,
+              type: 'info',
+              detail: 'locked',
+            });
+            loggedLocked = true;
+          }
+          // Wait a moment before attempting to acquire the lock again.
+          await new Promise((resolve) => setTimeout(resolve, 200));
```
justinfagnani, line 201: nit: might be nice to pull these three numeric constants here into named constants.

justinfagnani (approved):
Nice small PR for this!

src/test/util/uvu-timeout.ts
```diff
@@ -4,9 +4,13 @@
...
  */
 
+import {IS_WINDOWS} from '../../util/windows.js';
+
 import type * as uvu from 'uvu';
 
-const DEFAULT_TIMEOUT = Number(process.env.TEST_TIMEOUT ?? 30_000);
+const DEFAULT_TIMEOUT = Number(
+  process.env.TEST_TIMEOUT ?? (IS_WINDOWS ? 60_000 : 30_000)
+);
```
augustjk, line 13: Super nit: I'm guessing Windows specifically needed more time for some of these. Is there harm in just changing the default globally to 60? Or can we set the `TEST_TIMEOUT` environment variable for windows tests? The conditional default value in the code just seems kinda weird.

augustjk (approved):
This is awesome!!

README.md

src/executor.ts
aomarks: I dunno, sometimes I feel like that splits apart code that makes more sense together. Why is it better to need to look somewhere else in the file to find out what the number is?

src/test/parallelism.test.ts
aomarks: This is testing the `output=[]` case, when there is no lock. This test would fail if we did lock, because the 2nd invocation couldn't otherwise start before the 1st one ended.

Added comment.

src/test/util/uvu-timeout.ts
aomarks: Yeah, one of these needed more time on Windows. I've seen some other tests occasionally time out on windows too, so I figured I'd just do it universally -- but you're right it's probably better to keep the other ones at 30 seconds, since that gives faster results when there are genuine test failures. And we can just add more overrides the next time we see a timeout.

Issue #66, opened 2022-04-01
A way to specify options for all scripts in a package


aomarks:
Possibility:

```json
{
  "*": {
    "dependencies": [
      "bootstrap"
    ],
    "packageLocks": [
      "yarn.lock"
    ]
  }
}
```

`*` is unlikely to collide with a real script name, but we should probably support an escaping scheme anyway, e.g. `\*` means the script literally called `*`.

mihar-22:
This would be amazing, currently repeating `pnpm-lock.yaml` for every script is fragile and repetitive.

PR #193, opened 2022-05-06
Set --enable-source-maps when testing.

rictic:


aomarks (changes requested):
I believe this is built in to Node now with the [`--enable-source-maps`](https://nodejs.org/api/cli.html#--enable-source-maps) flag, so we can instead put:

`NODE_OPTIONS=--enable-source-maps` in front of each test command.

rictic:
Wow, it's built into node? That's great!

aomarks:
> Wow, it's built into node? That's great!

Ugh, it crashes Node 14 because the flag isn't recognized, and Windows because Windows has a different syntax for environment variables.

Maybe we just set this variable using `env` at the top of the GitHub Actions workflow (unless Node 14), and then we personally just put it in our `.bashrc` files?

aomarks:
> Ugh, it crashes Node 14 because the flag isn't recognized, and Windows because Windows has a different syntax for environment variables.
> 
> Maybe we just set this variable using `env` at the top of the GitHub Actions workflow (unless Node 14), and then we personally just put it in our `.bashrc` files?

Alternatively, we could use https://github.com/kentcdodds/cross-env which works on Windows.

I wonder if this should be a Wireit feature, it would be very easy to implement something like `{"env": {"foo":"bar"}}`.

aomarks:
> I wonder if this should be a Wireit feature, it would be very easy to implement something like `{"env": {"foo":"bar"}}`.

Would pair well with https://github.com/google/wireit/issues/66

aomarks (approved):

PR #194, opened 2022-05-06
Pull file-level analysis out into its own object.

rictic:
This way we can generate fewer duplicate diagnostics and walk the package.json file's large scale structure only once rather than once for each script.

Part of working towards IDE support, because what we _really_ want to do is say "give me all the diagnostics in these open files" or perhaps "starting from these files, give me all reachable diagnostics". However I think even just by itself this will be useful for deduplicating diagnostics.

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -27,6 +27,165 @@ import {Failure} from './event.js';
  */
 export type PlaceholderConfig = ScriptReference & Partial<ScriptConfig>;
 
+interface ScriptSyntaxInfo {
+  name: string;
+  /** The node for this script in the scripts section of the package.json */
+  scriptNode?: NamedAstNode<string>;
+  /** The node for this script in the wireit section of the package.json */
+  wireitConfigNode?: NamedAstNode;
+}
+/**
+ * A parsed and minimally analyzed package.json file.
+ *
+ * This does some very basic syntactic analysis of the package.json file,
+ * finding issues like "the scripts section isn't an object mapping strings to
+ * strings" and "the wireit section isn't an object mapping strings to objects".
+ *
+ * Makes it easy to find the syntax nodes for a script.
+ *
+ * Does not do any validation or analysis of the wirit script configs.
+ *
+ * This class exists in part so that we walk the package.json file only once,
+ * and in part so that we generate file-level syntactic diagnostics only once,
+ * so that we can do better deduplication of errors.
+ */
+class PackageJson {
+  readonly jsonFile: JsonFile;
+  // We keep the file level AST node private to represent the invariant that
+  // we only walk the file once, in this class, and nowhere else.
+  readonly #fileAstNode: JsonAstNode;
+  readonly #scripts: Map<string, ScriptSyntaxInfo> = new Map();
+  readonly failures: readonly Failure[];
+  readonly scriptsSection: NamedAstNode | undefined = undefined;
+  readonly wireitSection: NamedAstNode | undefined = undefined;
+  constructor(jsonFile: JsonFile, fileAstNode: JsonAstNode) {
+    this.jsonFile = jsonFile;
+    this.#fileAstNode = fileAstNode;
+    const failures: Failure[] = [];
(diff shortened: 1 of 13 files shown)


rictic:
Ah, got a windows path issue in an error message, will fix

rictic:
Ok, I think this is ready for review!

aomarks:
> This way we can generate fewer duplicate diagnostics and walk the package.json file's large scale structure only once rather than once for each script.

How did we generate duplicate diagnostics before? Wasn't everything always done on a per-script basis, and we only looked at each script once?

rictic:
Well the problem was threefold. First, some problems were file level, so every dependency would encounter it the same, second, execution problems were repeated for every dependent, and third there was a race condition in the caching package json reader such that even with some deduplication added we would still see dupes because caching a file was racy

src/util/package-json-reader.ts
```diff
@@ -4,35 +4,28 @@
...
 import {parseTree} from './ast.js';
 
-import type {PlaceholderConfig} from '../analyzer.js';
-import type {JsonAstNode} from './ast.js';
-import {Failure} from '../event.js';
-import {Result} from '../error.js';
-
 export const astKey = Symbol('ast');
 
 export interface JsonFile {
```
aomarks, line 16: Should `JsonFile` be in `ast.ts`? It's not used by `package-json-reader.ts` anymore.

src/analyzer.ts
```diff
@@ -97,6 +101,10 @@ export class Analyzer {
     return {ok: true, value: rootConfig};
   }
 
+  async #readPackageJson(packageDir: string): Promise<Result<PackageJson>> {
```
aomarks, line 104: Maybe just inline this, since it's only used in one place and is trivial and already readable.

src/analyzer.ts
```diff
@@ -908,31 +833,25 @@ const assertArray = (
  * Throw an error if it was an object literal ({...}), assuming it was parsed
```
aomarks, line 833: Can be another PR, but just noticed a bunch of these functions now have incorrect jsdoc, since they don't throw anymore.

src/event.ts
```diff
@@ -85,6 +85,10 @@ interface ErrorBase<T extends PackageReference> extends EventBase<T> {
   type: 'failure';
 }
 
+interface ScriptlessError {
```
aomarks, line 88: Isn't there always at least a package json path? Why can't `PackageJsonPathError` and `InvalidConfigSyntax` have `extends ErrorBase<PackageReference>>`?

src/logging/default-logger.ts
```diff
@@ -55,17 +55,21 @@ export class DefaultLogger implements Logger {
...
-    const label = this.#label(event.script);
-    const prefix = label !== '' ? ` [${label}]` : '';
     switch (type) {
       default: {
         throw new Error(`Unknown event type: ${unreachable(type) as string}`);
       }
 
       case 'success': {
         const reason = event.reason;
+        const prefix = this.#getPrefix(event.script);
```
aomarks, line 72: Related to other comment, not sure I understand why we wouldn't always have a `PackageReference` at least.

aomarks:
> Well the problem was threefold. First, some problems were file level, so for every script that we would attempt to run in that file we would generate the same failure. See the new test `"we don't produce a duplicate error when there's multiple deps into the same invalid file"`. That's a fairly rare case but it's there.
> 
> Second, failures during execution were logged for every dependent. I noticed this when hacking on wireit, which got me started down this path. Add a TypeScript error and run `npm test` and you'll see a log about how the build script exited with code 1 for each test script.
> 
> And third, there was a race condition in the caching package json reader such that even after adding deduplication logic we would still get dupes because caching a file was racy, and in a lot of situations we've probably been reading and parsing the same file many times. That's a very easy mistake to make, so I added a little async cache class with a safe API.

Got it, thanks for explanation! Great improvements.

aomarks (changes requested):
Looks great! Just a few questions/comments.

PR #198, opened 2022-05-09
Factor out `ScriptChildProcess` class

aomarks:
Factor out a new `ScriptChildProcess` class from the`#executeCommandIfNeeded` method. This class represents a spawned child process.

This will be useful for [server mode](https://github.com/google/wireit/issues/33), because it gives us a more discrete object that represents a running child process. The lifecycle of running processes will become more complex in server mode, especially in watch mode where we'll be persisting child processes potentially across different build graphs. In a follow up PR, it will gain a `state` property and a `terminate` method.

Also includes a small tweak to logging. We used to log the "Running" message before we tried to get a `WorkerPool` slot, which could give the impression that something was running that wasn't yet, when concurrency was constrained.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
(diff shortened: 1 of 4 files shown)


aomarks:
Sorry, just force pushed because I realized the incremental diff was actually more confusing than a full diff in this case. Ignore previous comment about looking at these one-by-ne.

rictic (approved):

augustjk (approved):

PR #199, opened 2022-05-09
Keep processing after the first error

rictic:
This is preparation for IDE work, where we'll still want to do as much analysis as possible.

As a side effect, we'll also see more diagnostics from the CLI, which may be helpful for figuring out the root cause of an issue rather than just stopping on the first one.

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -5,29 +5,65 @@
  */
 
 import * as pathlib from 'path';
-import {Diagnostic, MessageLocation, Result} from './error.js';
 import {CachingPackageJsonReader} from './util/package-json-reader.js';
-import {scriptReferenceToString} from './script.js';
 import {findNodeAtLocation, JsonFile} from './util/ast.js';
+import {scriptReferenceToString} from './script.js';
 
+import type {ArrayNode, JsonAstNode, NamedAstNode} from './util/ast.js';
+import type {Diagnostic, MessageLocation, Result} from './error.js';
+import type {DependencyOnMissingPackageJson, Failure} from './event.js';
+import type {PackageJson, ScriptSyntaxInfo} from './util/package-json.js';
 import type {
   ScriptConfig,
   ScriptReference,
   ScriptReferenceString,
 } from './script.js';
-import type {ArrayNode, JsonAstNode, NamedAstNode} from './util/ast.js';
-import {DependencyOnMissingPackageJson, Failure} from './event.js';
-import {PackageJson} from './util/package-json.js';
+
+/**
+ * A script config that might be at any point of the analysis pipeline,
+ * or have stalled at any point along that way due to errors.
+ */
+export type PotentiallyValidScriptConfig =
+  | UnvalidatedConfig
+  | LocallyValidScriptConfig
+  | ScriptConfig;
 
 /**
  * A {@link ScriptConfig} where all fields are optional apart from `packageDir`
  * and `name`, used temporarily while package.json files are still loading.
+ *
+ * A script with an invalid config may stay a placeholder forever.
  */
-export type PlaceholderConfig = ScriptReference & Partial<ScriptConfig>;
+export type UnvalidatedConfig = ScriptReference &
+  Omit<Omit<Partial<ScriptConfig>, 'state'>, 'dependencies'> & {
+    state: 'unvalidated';
+    failures: Failure[];
+    dependencies?: Array<
+      UnvalidatedConfig | LocallyValidScriptConfig | ScriptConfig
+    >;
+  };
+
+/**
+ * A ScriptConfig that is locally valid, but whose dependencies may not have
+ * resolved and checked for circular dependency errors yet.
+ */
+export type LocallyValidScriptConfig = Omit<
+  Omit<ScriptConfig, 'state'>,
+  'dependencies'
+> & {
+  state: 'locally-valid';
+  dependencies: Array<
+    UnvalidatedConfig | LocallyValidScriptConfig | ScriptConfig
+  >;
+};
 
 interface PlaceholderInfo {
-  placeholder: PlaceholderConfig;
-  failuresPromise: Promise<Result<void, Failure[]>>;
+  placeholder: PotentiallyValidScriptConfig;
+  /**
+   * A promise that resolves when this placeholder has either been upgraded to a
+   * LocallyValidScriptConfig, or if that process has failed.
+   */
+  upgradeComplete: Promise<undefined>;
 }
 
 /**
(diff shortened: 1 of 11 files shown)


src/executor.ts
```diff
@@ -362,9 +362,16 @@ class ScriptExecution {
     // Note we use Promise.allSettled instead of Promise.all so that we can
     // collect all errors, instead of just the first one.
     const dependencyResults = await Promise.allSettled(
-      this.#script.dependencies.map((dependency) =>
-        this.#executor.execute(dependency)
-      )
+      this.#script.dependencies.map((dependency) => {
```
aomarks, line 365: It sort of feels like we should be able to enforce that an `Executor` only takes a config that is guaranteed to be valid from the `Analyzer`, so that we don't have to check this during execution?

src/analyzer.ts
```diff
@@ -73,45 +113,88 @@ export class Analyzer {
 
     // Note we can't use Promise.all here, because new promises can be added to
     // the promises array as long as any promise is pending.
-    const errors = [];
-    while (this.#placeholderUpgradePromises.length > 0) {
-      const result = await this.#placeholderUpgradePromises.shift();
-      if (result?.ok === false) {
-        errors.push(result.error);
+    await this.#waitForanalysisToComplete();
```
aomarks, line 116:
```suggestion
    await this.#waitForAnalysisToComplete();
```

src/analyzer.ts
```diff
@@ -73,45 +113,88 @@ export class Analyzer {
...
-    if (errors.length > 0) {
-      return {ok: false, error: errors};
-    }
 
     // We can safely assume all placeholders have now been upgraded to full
     // configs.
-    const rootConfig = rootPlaceholder as ScriptConfig;
+    const rootConfig = rootPlaceholder.placeholder;
+    if (rootConfig.state === 'unvalidated') {
+      throw new Error(
```
aomarks, line 128: Just checking, is the rule that we `throw` if it's an internal implementation error, because that's more convenient and in some cases can keep noise out of the type system?

src/analyzer.ts
```diff
@@ -73,45 +113,88 @@ export class Analyzer {
...
     );
     if (!cycleResult.ok) {
       return {ok: false, error: [cycleResult.error]};
     }
-    return {ok: true, value: rootConfig};
+    const validRootConfig = cycleResult.value;
+    return {ok: true, value: validRootConfig};
+  }
+
+  async getDiagnostics(): Promise<Set<Failure>> {
```
aomarks, line 143: Can this be `#private`?

aomarks (changes requested):
Diff is a bit hard to read now that the other PR merged because I'm seeing unrelated changes, I think it needs a merge/rebase.

src/analyzer.ts
rictic: Yeah exactly. An internal error should never happen, but we can't rule it out with the type system, so just in case we throw an informative error.

src/analyzer.ts
rictic: I think it might need to be later on but we can make it private for now.

src/analyzer.ts
rictic: Done

src/analyzer.ts
aomarks: I guess we can think of `throw` like you would think of `panic` in Go or Rust.

src/analyzer.ts
```diff
@@ -5,36 +5,77 @@
...
+  | LocallyValidScriptConfig
+  | ScriptConfig;
 
 /**
  * A {@link ScriptConfig} where all fields are optional apart from `packageDir`
  * and `name`, used temporarily while package.json files are still loading.
+ *
+ * A script with an invalid config may stay a placeholder forever.
+ */
+export type UnvalidatedConfig = ScriptReference &
```
aomarks, line 37: Needs another merge I think? I'm still seeing code from another PR in the diff.

aomarks: (Sorry ignore this particular comment, I got 2 PRs mixed up)

src/analyzer.ts
```diff
@@ -77,37 +113,61 @@ export class Analyzer {
...
+      }
+    }
+    return failures;
+  }
+
+  async #waitForAnalysisToComplete() {
+    while (this.#ongoingWorkPromises.length > 0) {
+      const promise = this.#ongoingWorkPromises[0];
+      await promise;
+      if (promise === this.#ongoingWorkPromises[0]) {
```
aomarks, line 167: Need a comment here? Why wouldn't we just shift before awaiting?

Also should we pop, since the order shouldn't matter, and as you pointed out in another PR pop is faster?

src/analyzer.ts
```diff
@@ -417,119 +319,348 @@ export class Analyzer {
...
+      !dependenciesErrored &&
+      !commandError
+    ) {
+      placeholder.failures.push({
+        type: 'failure',
+        reason: 'invalid-config-syntax',
+        script: placeholder,
+        diagnostic: {
+          severity: 'error',
+          message: `A wireit config must set at least one of "wireit" or "dependencies", otherwise there is nothing for wireit to do.`,
```
aomarks, line 344:
```suggestion
          message: `A wireit config must set at least one of "command" or "dependencies", otherwise there is nothing for wireit to do.`,
```

src/test/errors-analysis.test.ts
```diff
@@ -668,17 +666,14 @@ test(
...
     assertScriptOutputEquals(
       done.stderr,
-      String.raw`
-❌ ch${'\t'}\ ild${
-        pathlib.sep
-      }package.json:2:3 Script "mis${'\t'}\ sing" not found in the scripts section of this package.json.
-      "scripts": {}
-      ~~~~~~~~~
-❌ package.json:8:23 Cannot find script named "mis\t\\ sing" in package "${rig.resolve(
+      `
```
aomarks, line 669: The `String.raw` change got undone?

aomarks (changes requested):

src/analyzer.ts
rictic: Yeah, this is tricky, and a bit trickier with pop(), but agree that's the better method to call.

Alternatively we might represent it as a Set instead of an array so we can quickly remove any element, but that doesn't have as nice of an API for "give me an arbitrary element" as an array has.

src/test/errors-analysis.test.ts
rictic: Oh damn it. Fixed.

src/analyzer.ts
rictic: Done

src/analyzer.ts
```diff
@@ -60,10 +60,8 @@ export type LocallyValidScriptConfig = Omit<
 interface PlaceholderInfo {
   placeholder: PotentiallyValidScriptConfig;
   /**
-   * A promise that resolves when this placeholder (but not necessarily its
-   * dependencies) has been upgraded. Note that if failures were encountered
-   * along the way, (either on placeholder.failures or on reading its
-   * PackageJson file) then the upgrade may not have been completed.
+   * A promise that resolves when this placeholder has either been upgraded to a
+   * LocallyValidScriptConfig, or if that rocess has failed.
```
aomarks, line 64:
```suggestion
   * LocallyValidScriptConfig, or if that process has failed.
```

aomarks (approved):
just 1 typo

src/analyzer.ts
aomarks: Good explanation in comment ty

src/analyzer.ts
rictic: Done!

PR #200, opened 2022-05-09
Wireit website skeleton

aomarks:
Preview at https://google.github.io/wireit/

- Eleventy
- Dark + light theme
- No content in this PR. Preview has README content, which I think is a good starting point, but we'll want to evolve/expand significantly. I'll send the README content in a followup PR.
- Mobile nav works but is not great. It just squishes the TOC at the top. Will have a followup that adds a proper mobile menu.

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>


justinfagnani:
How's this going to be set up for GitHub pages? I see this is being merged into the `main` branch, but the pages are currently set to serve rom `gh-pages`. I think if you set pages to use `main` then you get to use `/docs` or `/` as the folder, but this creates a `/website` folder.

aomarks:
> How's this going to be set up for GitHub pages? I see this is being merged into the `main` branch, but the pages are currently set to serve rom `gh-pages`. I think if you set pages to use `main` then you get to use `/docs` or `/` as the folder, but this creates a `/website` folder.

There will be a GitHub Actions script that builds the website and pushes the output into the `gh-pages` branch.

justinfagnani (approved):

website/content/_static/wireit.svg
```diff
@@ -0,0 +1,10 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 200 55" fill="none" stroke="#000" stroke-width="2" stroke-linecap="round">
```
rictic, line 1: Consider instead copying from the version at the root of the repo?

website/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "@11ty/eleventy": "^1.0.1",
+    "@11ty/eleventy-plugin-syntaxhighlight": "^4.0.0",
+    "markdown-it": "^13.0.1",
+    "markdown-it-anchor": "^8.6.3",
+    "prism-themes": "^1.9.0"
+  },
+  "devDependencies": {
+    "@11ty/eleventy-navigation": "^0.3.3",
+    "@web/dev-server": "^0.1.31"
+  }
```
augustjk, line 44: Not sure if it really matters but what makes something a dependency vs devDependency in this context?

augustjk (approved):

rictic (approved):

website/content/_static/wireit.svg
aomarks: In this case I did the pure white version for dark theme. The outlined version doesn't look very good to me when small, because the detail is too fine.

rictic:
I like the workflow I came up with in https://github.com/webcomponents/custom-elements-everywhere to do automatic deployment with preview urls for every PR

website/package.json
aomarks: Good point, the distinction is meaningless here. Switched to all dependencies.

aomarks:
> I like the workflow I came up with in https://github.com/webcomponents/custom-elements-everywhere to do automatic deployment with preview urls for every PR

Oh yeah, I was going to ask you about that. Looks great, we should do the same!

Looks like the previews just accumulate forever:

https://github.com/webcomponents/custom-elements-everywhere/tree/gh-pages/preview/preview/preview

I guess that's fine. Could have an occasional cleanup job?

PR #202, opened 2022-05-09
Implement and test our first code action

rictic:
(in wireit proper at least!)

### DIFF
--- a/src/language-server.ts
+++ b/src/language-server.ts
@@ -71,9 +71,10 @@ function log(...values: unknown[]) {
     connection.console.log(message);
   }
 }
-if (false as boolean) {
-  log(`Log is useful when developing, even if we don't use it currently.`);
-}
+
+// So that we can just console.log and console.error as usual.
+console.log = log;
+console.error = log;
 
 const documents: TextDocuments<TextDocument> = new TextDocuments(TextDocument);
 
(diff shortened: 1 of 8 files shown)


src/language-server.ts
```diff
@@ -71,9 +71,10 @@ function log(...values: unknown[]) {
     connection.console.log(message);
   }
 }
-if (false as boolean) {
-  log(`Log is useful when developing, even if we don't use it currently.`);
-}
+
+// So that we can just console.log and console.error as usual.
```
aomarks, line 75: Does this mean that if this module gets imported, all `console.log` behavior changes for the whole process?

Is it possible to isolate that more cleanly so that there isn't a global side-effect?

src/test/codeactions.test.ts
```diff
@@ -0,0 +1,279 @@
...
+        ...converter.ideRangeToRange(e.range),
+      };
+    })
+  );
+}
+
+test('can refactor a script to use wireit', async ({rig}) => {
+  await assertCodeAction({
+    rig,
+    contentsWithPipe: {scripts: {test: "echo| 'test'"}},
```
aomarks, line 146: What is the significance of the pipe characters in this test file? I think I missed something.

aomarks (changes requested):
Looks good! Couple questions.

src/language-server.ts
rictic: This module is a binary. If it's imported, it will crash immediately unless it's been launched with a node RPC channel.

We could separate it out into a binary script and a library with the implementation if you think that'd be cleaner.

src/test/codeactions.test.ts
rictic: Ah yeah, added a comment. That's to indicate the location of the carat, where the user's text input cursor is when they request code actions.

src/language-server.ts
aomarks: Ah I see. Seems ok.

Maybe update the top of this file description to give a little more context? It's currently documented as though it were inside the extension folder, so that could help in the future.

src/test/codeactions.test.ts
aomarks: Oh I see, makes sense thanks.

aomarks (approved):

PR #203, opened 2022-05-10
Recover from GitHub Actions cache rate limit errors

aomarks:
We sometimes get 429 rate limit errors from the GitHub Actions Caching service. This PR is a quick fix which detects rate limit errors, makes them non-fatal, and then disables caching for the remainder of that Wireit process.

Just a quick fix, we could do some smarter things with backoff in the future, but I think this should cut down on build failures.


augustjk (approved):
LGTM

rictic (approved):

PR #209, opened 2022-05-10
Don't start new scripts after failure by default, add option for old behavior

aomarks:
### Before

Previously, if a script failed, a new script could still start as long as the failure didn't occur in its transitive dependencies.

For example in this graph, after `fail` fails, `bar` would not start, but `foo` would:

```
 main
  / \
 |   |
 v   v
foo  bar
      |
      v
     fail
```

### After

Now, by default, as soon as a failure occurs anywhere in the build graph, new scripts will not be started, but ones that are already running will be allowed to finish.

The old behavior is still useful, e.g. when you want to have a complete picture of which scripts are passing vs failing, so the old behavior can be enabled by setting:

```bash
WIREIT_FAILURES=continue
```

The default can be set explicitly with (not sure about this name, open to suggestions):

```bash
WIREIT_FAILURES=no-new
```

Fixes https://github.com/google/wireit/issues/29

### Follow up

In a follow-up PR, I am planning to add a `WIREIT_FAILURES=kill` mode, which will be even more aggressive, and terminate running scripts as soon as the first one fails. I think this is also useful in some situations, for when you want to be notified about a failure as soon as possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
-- Added `WIREIT_FAILURES` environment variable with the following options:
+- Added `WIREIT_FAILURES` environment variable that controls what happens when a
+  script fails (meaning it returned with a non-zero exit code) with the
+  following options:
 
   - `no-new` (default): Allow running scripts to continue, but don't start new
     ones.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,21 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Added
+
+- Added `WIREIT_FAILURES` environment variable with the following options:
+
+  - `no-new` (default): Allow running scripts to continue, but don't start new
+    ones.
+  - `continue`: Allow running scripts to continue, and start new ones as long as
+    all of their dependencies succeeded.
+
+### Changed
+
+- Default failure mode changed from `continue` to `no-new` (see above for
+  definitions).
 
 ## [0.4.1] - 2022-05-10
 
--- a/README.md
+++ b/README.md
@@ -381,8 +381,9 @@ The benefit of Wireit's watch mode over built-in watch modes are:
 
 ## Failures and errors
 
-By default, when a failure occurs in any script, all scripts that are already
-running are allowed to finish, but new scripts are not started.
+By default, when a script fails (meaning it returned with a non-zero exit code),
+all scripts that are already running are allowed to finish, but new scripts are
+not started.
 
 In some situations a different behavior may be better suited. There is an
 additional mode, which you can set with the `WIREIT_FAILURES` environment
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@
   - [GitHub Actions caching](#github-actions-caching)
 - [Cleaning output](#cleaning-output)
 - [Watch mode](#watch-mode)
+- [Failures and errors](#failures-and-errors)
 - [Package locks](#package-locks)
 - [Recipes](#recipes)
   - [TypeScript](#typescript)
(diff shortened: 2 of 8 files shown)


src/test/basic.test.ts
```diff
@@ -1040,4 +1040,260 @@ test(
...
+      // on that. Wait a moment first to ensure Wireit unblocks `cancel` before
+      // `fail`.
+      await new Promise((resolve) => setTimeout(resolve, 50));
+      failParentBlockerInv.exit(0);
+
+      assert.equal((await wireit.exit).code, 1);
+      assert.equal(fail.numInvocations, 1);
+      assert.equal(failParentBlocker.numInvocations, 1);
+      assert.equal(cancel.numInvocations, 0);
+      assert.equal(cancelBlocker.numInvocations, 1);
```
augustjk, line 1155: It's not exactly the target of the test but should this also include `assert.equal(failParent.numInvocations, 0);` ?

augustjk (approved):
Looks great! Probably not blockers but just pointing out a couple things.

src/test/basic.test.ts
aomarks: failParent doesn't have a `command`, so there's nothing to verify for it

src/executor.ts
```diff
@@ -383,6 +477,12 @@ class ScriptExecution {
     }
 
     return this.#workerPool.run(async (): Promise<Result<void>> => {
+      // Significant time could have elapsed since we last checked because of
+      // parallelism limits.
+      if (this.#shouldCancel) {
+        return {ok: false, error: this.#cancelledEvent};
```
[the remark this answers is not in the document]
aomarks: Some of the sites are places where we return `ExecutionResult`s, which are `Result<ScriptState, Failure[]>`. That's the top-level thing we return from execution, and it's got an array of `Failure`s because when there are multiple dependencies, you might have multiple errors.

Other sites are more drilled-in functions, which just return a specific single error if they fail. Actually the `#acquireLock` one should probably be like that, done.

src/executor.ts
```diff
@@ -44,15 +54,27 @@ export class Executor {
...
+    cache: Cache | undefined,
+    failureMode: FailureMode
   ) {
     this.#logger = logger;
     this.#workerPool = workerPool;
     this.#cache = cache;
+    this.#failureMode = failureMode;
+  }
+
+  notifyFailure(): void {
```
rictic, line 72:
```suggestion
  /** 
    * Call this function when any script has failed to notify other scripts, and
    * potentially cancel/kill them based on the FailureMode.
    */
  notifyFailure(): void {
```

src/test/basic.test.ts
```diff
@@ -1040,4 +1040,260 @@ test(
...
+    const continuesInv = await continues.nextInvocation();
+    continuesInv.exit(0);
+
+    assert.equal((await wireit.exit).code, 1);
+    assert.equal(fail.numInvocations, 1);
+    assert.equal(continues.numInvocations, 1);
+    assert.equal(continuesBlocker.numInvocations, 1);
+  })
+);
+
```
rictic, line 1298: Great tests!

rictic (approved):

CHANGELOG.md
```diff
@@ -6,7 +6,21 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Added
+
+- Added `WIREIT_FAILURES` environment variable with the following options:
```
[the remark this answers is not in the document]
aomarks: Done

README.md
```diff
@@ -378,6 +379,27 @@ The benefit of Wireit's watch mode over built-in watch modes are:
   simultaneously, such as build steps being triggered before all preceding steps
   have finished.
 
+## Failures and errors
+
+By default, when a failure occurs in any script, all scripts that are already
```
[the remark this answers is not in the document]

src/executor.ts
aomarks: Done (similar)

src/test/basic.test.ts
```diff
@@ -1040,4 +1040,260 @@ test(
   })
 );
 
+for (const envSetting of ['no-new', undefined]) {
+  test(
+    `don't start new script after unrelated failure when WIREIT_PARALLEL=${
```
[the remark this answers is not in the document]
aomarks: Oops! Done.

PR #211, opened 2022-05-11
Improve a diagnostic's location

rictic:
When a script has a wireit config but not an entry in the scripts section, we should squiggle under the wireit config name if present.

We also give it a different reason code, so that we don't supercede this error message with an error about a mising dependency, as that's confusing. The dependency is present more or less, it's just improperly set up, so it's more helpful to tell the user about that improper setup than to just say that we couldn't find the dep.


src/test/errors-analysis.test.ts
```diff
@@ -119,6 +119,37 @@ test(
   })
 );
 
+test.only(
```
aomarks, line 122: Remove only

aomarks (approved):

src/test/errors-analysis.test.ts
rictic: Eep! Done

Also added an eslint pass to prevent this in the future and verified that it flagged this instance.

aomarks: Nice

aomarks (approved):

PR #214, opened 2022-05-11
Active script termination

aomarks:
Previously, we depended on the fact that we didn't use [`detached`](https://nodejs.org/api/child_process.html#optionsdetached) when spawning scripts, so that we didn't need to do anything special when the user sent a `SIGINT` to Wireit (e.g. via Ctrl-C). However, this meant that we didn't have the ability to terminate just one script; either Wireit and all child processes were killed, or none where.

Now, we enable `detached` mode (except on Windows where it isn't needed and is actually bad), so that we can control termination of specific scripts ourselves, and add a `terminate` method to `ScriptChildProcess`.  This is a prerequisite for the next PR, which will add `WIREIT_FAILURES=kill` (https://github.com/google/wireit/issues/31).

Also includes refactoring to the `Executor` to more clearly handle the 3 related events: `failureOccured`, `cancelNewScripts`, and `terminateRunningScripts`.

Fixes https://github.com/google/wireit/issues/30

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -124,7 +124,7 @@ export class Executor {
    * A promise which resolves if we should terminate running scripts.
    */
   get shouldTerminateRunningScripts(): Promise<void> {
-    return this.#cancelNewScripts.promise;
+    return this.#terminateRunningScripts.promise;
   }
 
   async execute(script: ScriptConfig): Promise<ExecutionResult> {
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -33,7 +33,7 @@ import type {
 import type {Logger} from './logging/logger.js';
 import type {WriteStream} from 'fs';
 import type {Cache} from './caching/cache.js';
-import type {Failure, Cancelled} from './event.js';
+import type {Failure, StartCancelled} from './event.js';
 
 type ExecutionResult = Result<ScriptState, Failure[]>;
 
@@ -57,10 +57,10 @@ export class Executor {
 
   /** Resolves when the first failure occurs in any script. */
   readonly #failureOccured = new Deferred<void>();
-  /** Resolves when we decide that new scripts should be cancelled. */
-  readonly #cancelNewScripts = new Deferred<void>();
-  /** Resolves when we decide that running scripts should be terminated. */
-  readonly #terminateRunningScripts = new Deferred<void>();
+  /** Resolves when we decide that new scripts should not be started. */
+  readonly #stopStartingNewScripts = new Deferred<void>();
+  /** Resolves when we decide that running scripts should be killed. */
+  readonly #killRunningScripts = new Deferred<void>();
 
   constructor(
     logger: Logger,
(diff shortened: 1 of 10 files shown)


src/script-child-process.ts
```diff
@@ -119,10 +188,74 @@ export class ScriptChildProcess {
...
           resolve({ok: true, value: undefined});
         }
+        this.#state = 'stopped';
       });
     });
   }
 
+  /**
+   * Mark this child process for termination. On Linux/macOS, sends a `SIGINT`
+   * signal. On Windows, invokes `taskkill /pid PID /t /f`.
```
augustjk, line 198: Perhaps the comment after the first sentence belongs on the `#actuallyTerminate` method?

augustjk (approved):

src/script-child-process.ts
aomarks: I was thinking that since this is the public one, it's more useful to put the details there, so that you can see it when you hover-over from another module.

augustjk: Ooh, okay that makes sense!

src/executor.ts
```diff
@@ -54,19 +54,52 @@ export class Executor {
   readonly #logger: Logger;
   readonly #workerPool: WorkerPool;
   readonly #cache?: Cache;
-  readonly #failureMode: FailureMode;
-  readonly #failureDeferred = new Deferred<void>();
+
+  /** Resolves when the first failure occurs in any script. */
+  readonly #failureOccured = new Deferred<void>();
+  /** Resolves when we decide that new scripts should be cancelled. */
```
rictic, line 60: Here and elsewhere, seems like "should not start new scripts" is clearer than "should cancel new scripts"

src/logging/default-logger.ts
```diff
@@ -175,6 +175,10 @@ export class DefaultLogger implements Logger {
             // fairly noisy. Maybe in a verbose mode.
             break;
           }
+          case 'terminated': {
+            console.error(`💀${prefix} Terminated`);
```
rictic, line 179: I encounter the term 'killed' more often than 'terminated' when talking about ending an ongoing process

src/script-child-process.ts
```diff
@@ -35,18 +38,25 @@ const PATH_ENV_SUFFIX = (() => {
   return entries.slice(endOfNodeModuleBins).join(pathlib.delimiter);
 })();
 
+export type ScriptChildProcessState =
+  | 'starting'
+  | 'started'
+  | 'stopping'
```
rictic, line 44: 'stopping' could be a bit clearer, it means that we're trying to kill the process yeah?

rictic (approved):
A few terminological suggestions, and one substantive one

src/executor.ts
aomarks: Done. New terms: `#stopStartingNewScripts`, `StartCancelled` event.

src/logging/default-logger.ts
aomarks: Done. New terms: `#killRunningScripts`, `Killed` event.

src/script-child-process.ts
aomarks: Done. Renamed to `killing`.

src/script-child-process.ts
```diff
@@ -119,10 +188,74 @@ export class ScriptChildProcess {
...
+      // Windows doesn't have signals. Node ChildProcess.kill() sort of emulates
+      // the behavior of SIGKILL (and ignores the signal you pass in), but this
+      // doesn't end child processes. We have child processes because the parent
+      // process is cmd.exe or PowerShell.
+      // https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/taskkill
+      spawn('taskkill', [
+        '/pid',
+        this.#child.pid.toString(),
+        /* End child processes */ '/t',
+        /* Forceful */ '/f',
```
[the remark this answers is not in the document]
aomarks: Sounds reasonable, done, and will see if the tests pass.

(In the future we may want to support harder kills in some situations, e.g. if the user Ctrl-C's twice in a row. So then we'd do `\f` on Windows, and `SIGTERM` on posix.)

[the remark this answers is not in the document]
aomarks: Update, in https://github.com/google/wireit/pull/220, removing the `/f` flag prevented Wireit from actually killing the Node process we spawn in the tests. I can't say I understand why -- I suppose it could be that Node doesn't actually respond to the Windows kill signal in the same way it does `SIGINT` on posix -- but I think that means we need `/f` to be reliable.

PR #216, opened 2022-05-12
Add README content to website

aomarks:
This is not the final content that we should use even before publicizing the site, but it's a good starting point.

Also makes the website servable from any directory, instead of having to be built knowing what the directory is up-front, by using relative paths. Needed a custom filter for this, surprisingly. This lets us serve the same exact build from any of these mount points:

```
http://localhost:8080/
https://google.github.io/wireit/
https://google.github.io/wireit/previews/<pr>/
```


website/.eleventy.cjs
```diff
@@ -29,8 +30,26 @@ module.exports = function (eleventyConfig) {
     })
   );
 
+  /**
+   * Generate a relative path to the root from the given page URL.
+   *
+   * Useful when a template which is used from different directories needs to
+   * reliably refer to a path with a relative URL, so that the site can be
+   * served from different sub-directories.
```
justinfagnani, line 38: How are you serving the site locally? I've handled this before by keeping `pathPrefix` and using the `--base-path` flag for Web Dev Server.

aomarks: Everything now uses a relative path, so it works from any directory as long as the server redirects to append `/`

aomarks: More context, we want this so that we can serve PR previews using the technique discussed here: https://github.com/google/wireit/pull/200#issuecomment-1121688775

website/content/_includes/layout.njk
```diff
@@ -1,17 +1,18 @@
...
 <!doctype html>
 <html lang="en">
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>{{ title }} | Wireit</title>
-    <link rel="stylesheet" href="{{ '/style.css' | url }}">
-    <link rel="stylesheet" href="{{ '/prism-light.css' | url }}" media="(prefers-color-scheme:light)">
-    <link rel="stylesheet" href="{{ '/prism-dark.css' | url }}" media="(prefers-color-scheme:dark)">
+    <link rel="stylesheet" href="{{ base }}/style.css">
```
justinfagnani, line 8: No change needed now, but is is possible to write the filter so that the usage is the same as `url`? I find that a bit easier to read and follow.

justinfagnani (approved):

website/content/_includes/layout.njk
aomarks: It would look like:

```
{{ 'style.css' | relativePathToRoot(page.url) }}
```

Which I kinda find less readable, personally.

augustjk (approved):

PR #228, opened 2022-05-13
Release 0.4.2

aomarks:


augustjk (approved):
Not necessarily breaking but adds new API and changing default behavior. Would a minor bump not be more appropriate? Probably not a big deal at this stage of versioning.

aomarks:
> Not necessarily breaking but adds new API and changing default behavior. Would a minor bump not be more appropriate? Probably not a big deal at this stage of versioning.

I tend to follow the rule that when the major digit is `0`, then the minor digit is used for breaking changes, and the patch digit is used for new features and bugfixes. That seems good because most people will have a semver range like `^0.1.0` since that's the default when you do `npm i wireit`, and that means they won't automatically receive upgrades in either the major _or_ minor digits (can confirm [here](https://semver.npmjs.com/) by entering wireit +  `^0.1.0`), because the `^` operator starts from the left-most-non-zero-digit.

aomarks:
> > Not necessarily breaking but adds new API and changing default behavior. Would a minor bump not be more appropriate? Probably not a big deal at this stage of versioning.
> 
> I tend to follow the rule that when the major digit is `0`, then the minor digit is used for breaking changes, and the patch digit is used for new features and bugfixes. That seems good because most people will have a semver range like `^0.1.0` since that's the default when you do `npm i wireit`, and that means they won't automatically receive upgrades in either the major _or_ minor digits (can confirm [here](https://semver.npmjs.com/) by entering wireit + `^0.1.0`), because the `^` operator starts from the left-most-non-zero-digit.

Just realized you were probably referring to the default behavior change. I don't think that should be considered a breaking change, because it only affects failures, and if you got a failure you shouldn't be doing any work downstream of that failed build, either way.

augustjk:
> > > Not necessarily breaking but adds new API and changing default behavior. Would a minor bump not be more appropriate? Probably not a big deal at this stage of versioning.
> > 
> > 
> > I tend to follow the rule that when the major digit is `0`, then the minor digit is used for breaking changes, and the patch digit is used for new features and bugfixes. That seems good because most people will have a semver range like `^0.1.0` since that's the default when you do `npm i wireit`, and that means they won't automatically receive upgrades in either the major _or_ minor digits (can confirm [here](https://semver.npmjs.com/) by entering wireit + `^0.1.0`), because the `^` operator starts from the left-most-non-zero-digit.
> 
> Just realized you were probably referring to the default behavior change. I don't think that should be considered a breaking change, because it only affects failures, and if you got a failure you shouldn't be doing any work downstream of that failed build, either way.

Yup, that was it. Makes perfect sense. Thanks for elaborating!

PR #229, opened 2022-05-13
Upgrade our own version of Wireit

aomarks:


AndrewJakubowicz (approved):

augustjk:
Version in the `package.json` isn't explicitly updated though the selector means we're gonna grab the latest patch.

aomarks:
> Version in the `package.json` isn't explicitly updated though the selector means we're gonna grab the latest patch.

Yep, I think that's fine.

PR #231, opened 2022-05-14
[ide] Be slightly more generous with certain locations

rictic:
When we're looking for the meaning of a location in a file, instead of only counting the name and the value of a property, also count the colon and whitespace in between.

i.e. if you're trying to jump to the definition of a wireit config from the scripts section, currently you can do so from either:

```
    "a": "wireit"
    ~~~
```

or

```
    "a": "wireit"
         ~~~~~~~~
```

but after this change you can do it anywhere in:

```
    "a": "wireit"
    ~~~~~~~~~~~~~
```

### DIFF
--- a/src/test/ide.test.ts
+++ b/src/test/ide.test.ts
@@ -493,7 +493,7 @@ test('can jump from scripts section to wireit config', async ({rig}) => {
   });
 });
 
-test('can jump from from colon in scripts section to wireit config', async ({
+test('can jump from colon in scripts section to wireit config', async ({
   rig,
 }) => {
   const ide = new IdeAnalyzer();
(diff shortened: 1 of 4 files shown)


src/test/ide.test.ts
```diff
@@ -493,4 +493,43 @@ test('can jump from scripts section to wireit config', async ({rig}) => {
   });
 });
 
+test('can jump from from colon in scripts section to wireit config', async ({
```
aomarks, line 496:
```suggestion
test('can jump from colon in scripts section to wireit config', async ({
```

aomarks (approved):
Nice!

Issue #107, opened 2022-04-17
Improvements around @actions/cache


aomarks:
### Problem

The public API provided by the `@actions/cache` package doesn't exactly meet our needs, because it automatically uses the file paths that are included in the tarball as part of the cache entry version (see https://github.com/actions/toolkit/blob/7654d97eb6c4a3d564f036a2d4a783ae9105ec07/packages/cache/src/internal/cacheHttpClient.ts#L70), and implements globbing differently.

We want complete control over our cache key, instead of having it be generated automatically based on file paths -- and we want to be sure we are using identical globbing logic to the rest of Wireit.

### Current solution

For this reason, we are currently reaching into the `internal/` directory of `@actions/cache` to get more control. This is bad because those modules could change at any time, which is why we currently have a strict (`"="`) version pin in our package.json.

It's also why we currently have `"skipLibCheck": false` in our `tsconfig.json`, and why we have the file `types/action-cache-contracts.d.ts` -- because the file `lib/internal/contracts.d.ts` is missing from the published `@actions/cache` package.

The `@actions/cache` package is also our largest dependency by far. It's 22MB, and adds 63 transitive dependencies.

### Options

1. We could file an issue or send a PR that provides a way to directly specify the cache key in `@actions/cache`. This would solve the version pinning problem, the `"skipLibCheck": false` problem, and would allow us to remove `types/action-cache-contracts.d.ts` -- but we'd still have the large dependency.

2. We could potentially move all of this logic into a separate package which is installed only by `google/wireit@setup-github-actions-caching/v1` -- instead of the main package. The action would then spin up its own HTTP server, which we would talk to instead with a more minimal API (note that the server we spin up would have direct filesystem access, so _it_ could make the tarballs). This would shrink the main `wireit` package's dependencies and filesize back down again (though the dependencies would still be installed -- just only in CI, instead of also locally), and would have the added benefit of not requiring us to expose the `ACTIONS_CACHE_URL` and `ACTIONS_RUNTIME_TOKEN` variables to all `run` steps (see https://github.com/actions/toolkit/issues/1053).

5. The logic we need from `@actions/cache` could be re-implemented from scratch in a minimal way, such that we could drop the dependency on `@actions/cache` all together. The main tricky part is the way it handles tarball generation across platforms (https://github.com/actions/toolkit/blob/7654d97eb6c4a3d564f036a2d4a783ae9105ec07/packages/cache/src/internal/tar.ts).

PR #234, opened 2022-05-15
Replace GitHub cache library with our own implementation

aomarks:
Replaces [`@actions/cache`](https://github.com/actions/toolkit/tree/main/packages/cache) with our own implementation for interacting with the GitHub Actions cache service.

### API mismatch with @actions/cache

`@actions/cache` takes a key and a set of glob patterns, and the literal glob patterns themselves are automatically used to generate a cache "version". The key + version then serve as a compound key for the cache entry.

This didn't work well for us because the glob implementation it uses doesn't support exclusions the way we need, so it is incompatible with our glob behavior, and may also be different in other subtle ways. We really just want to provide a list of files/directories for the tarball, and set the key independently.

But if we list all the files directly, then we can never get a cache hit, because the full list of files needs to be part of the cache key, but knowing the list of files requires running the script, which is what we're trying to avoid by restoring from cache!

### The workaround we had before

The workaround we had before was to reach into the `internal/` directory of `@actions/cache` and call some of the lower level functions directly, which let us control exactly what the cache key + version was, while still passing an explicit list of files to `tar`.

This worked ok, but had a high risk of breakage, because it uses non-public APIs. https://github.com/google/wireit/pull/227 is an example of us breaking because of that. It also required us to write some of our own typings, and to turn off `skipLibCheck` in our `tsconfig.json`. It was still quite a lot of code, too.

In addition, we still had to deal with an edge case relating to empty directories. By default, `tar` includes all recursive contents of a directory -- so if we wanted to cache an empty directory that wasn't actually empty on disk, we had to resort to a weird "empty directories manifest" hack. But now we can just use `tar --no-recursion` flag which avoids that whole issue.

### This PR

This PR completely replaces `@actions/cache` with a custom implementation that uses `https.request` and `execFile('tar', ...)`.

As a big bonus, **our install size decreased from 25MB to 2.4MB** and our **transitive dependency count decreased from 93 to 29**.

### Testing

- In https://github.com/google/wireit/pull/232 I improved our fake in a number of ways to make it higher fidelity, such as validating headers, and handling multiple upload chunks.

- I read through the implementation details in https://github.com/actions/toolkit/tree/main/packages/cache to make sure we are compatible, picking the same chunk sizes, etc.

- https://github.com/google/wireit/actions/runs/2325580687 is a run that tests this PR against the live servers by installing from an `npm pack`ed version. https://github.com/google/wireit/actions/runs/2325616423 is a follow up re-run, to show that everything got restored from cache.

Fixes https://github.com/google/wireit/issues/107


src/caching/github-actions-cache.ts
```diff
@@ -184,123 +152,182 @@ export class GitHubActionsCache implements Cache {
...
+  }
 
-    const compressionMethod = await GitHubActionsCache.#compressionMethod;
-    const version = this.#computeVersion(stateStr, compressionMethod);
-
-    let emptyDirsManifestPath: string | undefined;
-    if (emptyDirs.size > 0) {
-      const emptyDirsManifest = JSON.stringify([...emptyDirs]);
-      emptyDirsManifestPath = this.#emptyDirectoriesManifestPath(
+  async #reserveUploadAndCommitTarball(
```
rictic, line 171: Document what the boolean return means

src/caching/github-actions-cache.ts
```diff
@@ -184,123 +152,182 @@ export class GitHubActionsCache implements Cache {
...
+      return false;
+    }
+    if (!(await this.#commit(script, id, tarballBytes))) {
+      return false;
+    }
+    return true;
+  }
 
-    const tarballPath = await this.#makeTarball([...files], compressionMethod);
+  async #upload(
```
rictic, line 217: Ditto

src/caching/github-actions-cache.ts
```diff
@@ -184,123 +152,182 @@ export class GitHubActionsCache implements Cache {
...
-      await tarballDeleted;
+      await tarballHandle.close();
+    }
+  }
+
+  async #commit(
+    script: ScriptReference,
+    id: number,
+    tarballBytes: number
+  ): Promise<boolean> {
```
rictic, line 282: ditto

rictic (approved):
Excellent comments, great work validating this

src/caching/github-actions-cache.ts
aomarks: Done

src/caching/github-actions-cache.ts
aomarks: Done

src/caching/github-actions-cache.ts
aomarks: Done

Issue #237, opened 2022-05-16
Option to stop inheritance of cache keys from dependencies


aomarks:
Currently the cache key for a script (which affects whether it can be skipped or restored-from-cache) automatically includes the transitive cache keys of all of its dependencies. That means if A depends on B, then if an input to B changes, then both B and A will re-run -- regardless of whether B produces any new or relevant output.

This is a safe and good default because it doesn't require you to specify your input files, which means you can start seeing benefits sooner, and makes it harder to create an invalid config at first.

The downside is that you sometimes get unnecessary builds, because sometimes a dependency runs, but doesn't produce new output -- or doesn't produce new output that you actually care about. For example, a change to a `.ts` file might only cause changes to `d.ts` files, which Rollup can typically ignore. Or a change to a `tsconfig.json` type-checking rule might not produce any different output at all.

We should add a way to annotate that a particular dependency edge doesn't inherit the cache key.

Issue #23, opened 2022-03-22
Concise way to depend on scripts in child and dependency workspaces


aomarks:
#### [1] Run `<script>` in all of my workspaces (parent → child).

With Wireit, you can already just do `npm run build -ws` to run a given script in all workspaces, which is the standard npm workspaces approach. However it's not fully optimal, because npm doesn't parallelize. So we will also support a syntax like this:

```json
{
  "name": "root",
  "scripts": {
    "build": "wireit"
  },
  "wireit": {
    "build": {
      "dependencies": [
        {
          "script": "build",
          "packages": "workspaces"
        }
      ]
    }
  },
  "workspaces": [
    "packages/foo",
    "packages/bar"
  ]
}
```

Which is equivalent to:

```json
{
  "name": "root",
  "scripts": {
    "build": "wireit"
  },
  "wireit": {
    "build": {
      "dependencies": {
        [
          "./packages/foo:build",
          "./packages/bar:build"
        ]
      }
    }
  }
}
```

#### [2] Run `<script>` in all of my dependencies (child → siblings).

Related, it is often useful to run some script in all of the current packages dependencies (where those dependencies are workspaces contained by the same workspace root).

```json
{
  "name": "foo",
  "scripts": {
    "build": "wireit"
  },
  "wireit": {
    "build": {
      "dependencies": [
        {
          "script": "build",
          "packages": "dependencies"
        }
      ]
    }
  },
  "dependencies": {
    "bar",
    "baz"
  }
}
```

Which is equivalent to:

```json
{
  "name": "foo",
  "scripts": {
    "build": "wireit"
  },
  "wireit": {
    "build": {
      "dependencies": [
        "../bar:build",
        "../baz:build"
      ]
    }
  }
}
```

appsforartists:
I'm faking this in the interim by adding this to the root `package.json`

```json
  "scripts": {
    "build": "wireit"
  },
  "wireit": {
    "build": {
      "dependencies": [
        "./packages/core:build",
        "./packages/preact:build"
      ]
    }
  },
```

where `core` and `preact` are the folder names of my local packages.  Now I can, e.g. `yarn run build watch` and have all my packages transitively watched without running into serial blocking issues.

black7375:
IMO, I hope it can be set like t[urborepo's `dependsOn`](https://turborepo.org/docs/core-concepts/pipelines#task-dependency-format).

rictic:
What would "^build" in a dependency mean?

[...]

PR #238, opened 2022-05-16
Add triggersRerun setting to control fingerprint inheritance

aomarks:
### Background

Currently, if script A depends on script B, then script B's fingerprint is automatically included in script A's fingerprint. That means if an input file for script B changes, which causes B to re-run, then script A will re-run too — _regardless of whether script B actually emitted different output._

This is a good and safe default because it means you aren't required to specify the input files for every script when those input files are generated by a dependency; Wireit assumes that any time script B runs, script A could be affected. However, it comes with the trade-off that scripts will sometimes re-run even when none of their input files changed.

### This PR

This PR adds a `"triggersRerun": false` setting that can be annotated on dependencies. This prevents the fingerprint from being inherited, so running B won't *necessarily* cause A to run as well.

This can be used for more optimal builds via fewer re-runs, but requires that input files are always fully specified.

This will also be very useful for [service mode](https://github.com/google/wireit/issues/33), because a `triggersRerun:false` dependency won't require a restart across watch iterations. (Restarts will happen only if the fingerprint is different between iterations). For example, the script that builds the service itself can be a regular dependency (causing restarts), but the script that builds the *assets* served by the script can be a `triggersRerun:false` dependency (not causing restarts).

Note in order to have a place to put this annotation, dependencies can now be objects. They can still be plain strings as well, but must be objects to receive annotations. (This also opens the door for other annotations on dependency edges we may want in the future. E.g. https://github.com/google/wireit/issues/23 is probably better expressed as a `"workspaces": true` annotation instead of a magic `$WORKSPACES:` prefix as previously planned).

Fixes https://github.com/google/wireit/issues/237

#### Example

In this example, `bundle` has a `triggersRerune:false` dependency on `build`. Importantly, it also includes `lib/**/*.js` in its `files` array, which are the specific outputs from `tsc` that `rollup` consumes. Including these input files wasn't necessary before, but with `triggersRerun:false` it is now critical.

The advantage of this configuration is that if `tsc` re-runs but doesn't produce different `.js` files (for example, if it only produced different `.d.ts` files), then `rollup` won't need to re-run. We've also excluded the `lib/test` directory, because we know test files aren't included in our bundles.

```json
{
  "scripts": {
    "build": "wireit",
    "bundle": "wireit"
  },
  "wireit": {
    "build": {
      "command": "tsc",
      "files": ["src/**/*.ts", "tsconfig.json"],
      "output": ["lib/**"]
    },
    "bundle": {
      "command": "rollup -c",
      "dependencies": [
        {
          "script": "build",
          "triggersRerun": false
        }
      ],
      "files": ["rollup.config.json", "lib/**/*.js", "!lib/test"],
      "output": ["dist/bundle.js"]
    }
  }
}
```


src/analyzer.ts
```diff
@@ -481,17 +481,99 @@ export class Analyzer {
...
+          packageJson.jsonFile
+        );
+        if (!specifierResult.ok) {
+          encounteredError = true;
+          placeholder.failures.push(specifierResult.error);
+          continue;
+        }
+        const softResult = findNodeAtLocation(maybeUnresolved, ['soft']);
+        if (softResult !== undefined) {
+          if (softResult.value === true || softResult.value === false) {
```
augustjk, line 532: Super nit, maybe not worth it, but perhaps?
```suggestion
          if (typeof softResult.value === 'boolean') {
```

src/analyzer.ts
```diff
@@ -481,17 +481,99 @@ export class Analyzer {
...
+          packageJson.jsonFile
+        );
+        if (!specifierResult.ok) {
+          encounteredError = true;
+          placeholder.failures.push(specifierResult.error);
+          continue;
+        }
+      } else if (maybeUnresolved.type === 'object') {
+        specifierResult = findNodeAtLocation(maybeUnresolved, ['script']);
+        if (specifierResult == null) {
```
augustjk, line 501: Assuming this is specifically `==` to catch both `undefined` and `null`?

augustjk (approved):

aomarks:
In a discussion @rictic and I had yesterday, we mused over names that might be better than "soft" for this feature.

We considered some names that tried to encode more meaning, things like:

- `output-in-input-files`
-`inherits-fingerprint`
- `always-runs`

But we had trouble thinking of one that felt really clear. One way these names can be confusing is that it can be ambiguous what the directionality is (e.g. is it the dependency that always runs, or the dependee?)

I think we agreed that there might actually be a benefit to using a fuzzy term like "soft", because it basically requires the user to look at the actual definition (in the docs or by hovering if they have the vscode extension), rather than giving them a false impression.

@rictic suggested "weak" as an alternative to "soft", which I prefer.

Any other thoughts from reviewers here?

src/analyzer.ts
aomarks: I copied this style from the other code @rictic wrote in this file. I have been in the habit of using `=== undefined` when we know we don't need to handle `null` (a habit I learned from @justinfagnani) -- but maybe @rictic has a reason to use `== null` here?

src/test/ide.test.ts
```diff
@@ -322,6 +322,40 @@ test(`we jump to the scripts section for a vanilla script`, async ({rig}) => {
...
+    expected: {
+      target: `
+    "b": "echo"
+    ~~~~~~~~~~~`,
+      targetSelection: `
+    "b": "echo"
+    ~~~`,
+      // TODO(aomarks) The ~~~ is 2 spaces ahead of where it should be.
+      originSelection: `
+          "script": "b"
```
rictic, line 353: Oh, I bet I know what this is. The comparison that we use ignores leading and trailing whitespace. It probably should just ignore leading newlines. I bet if you add a couple of spaces to this line it will fix it

rictic (approved):

rictic:
We might want to release a new version of the extension before we release wireit, so that by the time people update wireit they'll probably already have the version extension that supports object deps

rictic:
Ultimately, I think that soft is a little better than weak. Weak in the sense of WeakMap or WeakRef suggests to me that the _dependency_ might not run, rather than the _dependent_.

Soft doesn't have that association so I think it's slightly better.

augustjk:
I like `soft` over `weak` for reasons @rictic mentioned above.

Something that occurred to me though, would there be any benefit to having a script level option that treats all of its dependency as `soft`? That would be something like `always-use-files-for-freshness` or `file-only-fingerprint` kind of thing?

Since putting even just one dependency that is soft means the user must be sure to specify a  comprehensive `files` input for the script, I wonder if it makes more sense that it's at the script level, rather than per dependency.

aomarks:
> Something that occurred to me though, would there be any benefit to having a script level option that treats all of its dependency as `soft`? That would be something like `always-use-files-for-freshness` or `file-only-fingerprint` kind of thing?
> 
> Since putting even just one dependency that is soft means the user must be sure to specify a comprehensive `files` input for the script, I wonder if it makes more sense that it's at the script level, rather than per dependency.

If a dependency is soft, you only need to be sure to specify the input files you care about from that one dependency. And there are legitimate cases where you want some dependencies soft, and some not.

I think allowing you to set it for a whole script could be a little dangerous, because you might set that flag, and then later add a dependency without actively considering whether it's safe for that dependency to be soft. With it being explicit per dependency, you're more likely to think about it on a per-dependency basis.

src/test/ide.test.ts
aomarks: Oh yeah that was it, done.

src/analyzer.ts
justinfagnani: I use `== null` to signify that either we do need to handle both (good), or that I'm not sure (bad - and I leave a TODO).

src/executor.ts
```diff
@@ -682,17 +679,24 @@ class ScriptExecution {
...
   async #computeState(
-    dependencyStates: Array<[ScriptReference, ScriptState]>
+    dependencyStates: Array<[Dependency, ScriptState]>
   ): Promise<ScriptState> {
     let allDependenciesAreCacheable = true;
     const filteredDependencyStates: Array<
       [ScriptReferenceString, ScriptState]
     > = [];
     for (const [dep, depState] of dependencyStates) {
+      if (dep.soft) {
```
justinfagnani, line 689: 👍

justinfagnani (approved):

src/analyzer.ts
aomarks: Switched everything to check `=== undefined` when the type indicated that it could not be null (which is almost everything).

aomarks:
I had a discussion with @justinfagnani about this change, and I've written up a draft proposal for an alternative way to think about this change: https://github.com/google/wireit/issues/245

aomarks:
This PR is ready for another review.

- I've merged in the new [services](https://github.com/google/wireit/pull/508) changes. For services, this setting has the effect whether the service needs to *restart* or not in watch mode when that particular dependency changes.

- Added tests and documentation about how this setting interacts with services.

- I've renamed `soft:true` to `triggersRerun:false`. Although it's more verbose, I think it gives a better impression of what the setting does.


  I would still like to pursue the idea of "output slices" described in https://github.com/google/wireit/issues/245 later, but I think we can still land this in the meantime. If we do add output slices, then that will be an additional, more natural way to express the same thing.

  However, I think this setting will still have a place -- [1] for services that read output dynamically (where it's not a question of which output you consume, it's more about whether the binary reads it at startup time or request time), and [2] because maybe sometimes it won't make sense to have your dependency declare an output slice (maybe it's just weirdly specific, or maybe you don't *own* that other script).

cc @justinfagnani

justinfagnani (approved):

PR #250, opened 2022-05-20
Allow file-only scripts, and disallow output without command

aomarks:
1. It is now possible to define a script that only defines input files. This can be useful for organizing groups of shared input files that multiple scripts depend on, such as configuration files. We've already been using this pattern in the Lit repo, but needed the workaround of setting a no-op `true` or `echo` command so that analysis passed.

2. Setting `"output"` on a script that does not have a `"command"` is now an error. It was possible before, and would even trigger caching, but that never made sense, because it couldn't have been that script that produced the output.

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -402,12 +402,15 @@ export class Analyzer {
       }
     }
 
+    const files = this.#processFiles(placeholder, packageJson, syntaxInfo);
+
     if (
       wireitConfig !== undefined &&
-      command === undefined &&
       dependencies.length === 0 &&
       !dependenciesErrored &&
-      !commandError
+      command === undefined &&
+      !commandError &&
+      (files === undefined || files.values.length === 0)
     ) {
       placeholder.failures.push({
         type: 'failure',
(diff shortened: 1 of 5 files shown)


augustjk (approved):
🎉

src/analyzer.ts
```diff
@@ -402,20 +402,23 @@ export class Analyzer {
...
+
     if (
       wireitConfig !== undefined &&
-      command === undefined &&
       dependencies.length === 0 &&
       !dependenciesErrored &&
-      !commandError
+      command === undefined &&
+      !commandError &&
+      (files === undefined || files.values.length === 0)
```
AndrewJakubowicz, line 413: This PR is so elegant and minimal  I'm doing a quick clarification. Checking my understanding that you've allowed file-only scripts by making the error more permissive. And it "just works".

Nice.

AndrewJakubowicz (approved):

src/analyzer.ts
aomarks: Yep!

PR #252, opened 2022-05-22
Misc refactoring of OneShotExecutor

aomarks:
⚠️ **Note to reviewer**: I recommend reading commits 1-by-1.

Some refactoring to make `OneShotExecution` easier to follow.

There are 3 outcomes for an execution: 1. fresh, 2. stale with a cache hit, 3. stale and we run the script. Previously handling for the 3 cases were sort of mixed up together in the `#execute` method, so it was a little hard to follow what happened in each case. Now the `#execute` method delegates to a distinct method for each outcome (`#handleFresh`, `#handleCacheHit`, `#handleNeedsRun`).

Also:

- Moved unexpected exception handling into the `Executor`, which simplifies dependency handling so that we don't have to use `Promise.allSettled`.

- Removed some unnecessary abort checks. `NoOp` executors don't need to worry about aborting at all, because so little time is spent in them, so the upstream or downstream scripts will handle aborts just fine. Similarly, `OneShotExecution` scripts also don't need to check for aborts before traversing dependencies, because we are doing a depth-first walk, so we will hit the leafs immediately anyway.

- Made the API for acquiring system-wide locks take a callback instead of returning a release function. More foolproof, and more similar to the `WorkerPool` API.

- Made a `Fingerprint` class which handles fingerprint creation plus lazy conversion between object and string forms. Gets rid of a bunch of `fingerprintStr` vs `fingerprintData` sort of variables we had to track before. Also added a brand to make it harder to directly create a `FingerprintData` without using the class.

### DIFF
--- a/src/execution/one-shot.ts
+++ b/src/execution/one-shot.ts
@@ -80,7 +80,8 @@ export class OneShotExecution extends BaseExecution<OneShotScriptConfig> {
       // Note we must wait for dependencies to finish before generating the
       // cache key, because a dependency could create or modify an input file to
       // this script, which would affect the key.
-      const fingerprint = await this.computeFingerprint(
+      const fingerprint = await Fingerprint.compute(
+        this.script,
         dependencyFingerprints.value
       );
       if (await this.#fingerprintIsFresh(fingerprint)) {
--- a/src/execution/one-shot.ts
+++ b/src/execution/one-shot.ts
@@ -67,9 +67,6 @@ export class OneShotExecution extends BaseExecution<OneShotScriptConfig> {
   }
 
   async #execute(): Promise<ExecutionResult> {
-    if (this.#shouldNotStart) {
-      return {ok: false, error: [this.#startCancelledEvent]};
-    }
     const dependencyFingerprints = await this.executeDependencies();
     if (!dependencyFingerprints.ok) {
       dependencyFingerprints.error.push(this.#startCancelledEvent);
--- a/src/execution/one-shot.ts
+++ b/src/execution/one-shot.ts
@@ -515,23 +515,30 @@ export class OneShotExecution extends BaseExecution<OneShotScriptConfig> {
   }
 
   /**
-   * Read this script's fingerprint file.
+   * Read this script's previous fingerprint from `fingerprint` file in the
+   * `.wireit` directory. Cached after first call.
    */
   async #readPreviousFingerprint(): Promise<Fingerprint | undefined> {
-    try {
-      return Fingerprint.fromString(
-        (await fs.readFile(
-          this.#fingerprintFilePath,
-          'utf8'
-        )) as FingerprintString
-      );
-    } catch (error) {
-      if ((error as {code?: string}).code === 'ENOENT') {
-        return undefined;
-      }
-      throw error;
+    if (this.#cachedPreviousFingerprint === undefined) {
+      this.#cachedPreviousFingerprint = (async () => {
+        try {
+          return Fingerprint.fromString(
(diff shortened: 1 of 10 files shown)


src/fingerprint.ts
```diff
@@ -10,7 +10,7 @@ import type {ScriptReferenceString} from './script.js';
  * All meaningful inputs of a script. Used for determining if a script is fresh,
  * and as the key for storing cached output.
  */
-export interface Fingerprint {
+export interface FingerprintData {
```
rictic, line 13: It occurs to me that the order of fields matters in FingerprintData, since they're compared by JSON serialization. I think that should be documented here.

It may be worthwhile to more tightly control construction. For example by adding a brand on the type, as syntactic vinegar to make a speed bump when constructing a FingerprintData.

src/execution/one-shot.ts
```diff
@@ -515,23 +515,30 @@ export class OneShotExecution extends BaseExecution<OneShotScriptConfig> {
...
+          if ((error as {code?: string}).code === 'ENOENT') {
+            return undefined;
+          }
+          throw error;
+        }
+      })();
     }
+    return this.#cachedPreviousFingerprint;
   }
+  #cachedPreviousFingerprint?: Promise<Fingerprint | undefined>;
```
rictic, line 577: Do we use standard class fields? If not, consider an explicit init to undefined

src/fingerprint.ts
aomarks: Moved creation into the fingerprint class itself and added a brand, so that it's hard to create a fingerprint except through that class.

src/execution/one-shot.ts
aomarks: We use down-leveled standard class fields (though maybe we shouldn't given the extra book-keeping, it's because we originally emitted native class fields, but then downgraded to Node 14 support).

src/fingerprint.ts
```diff
@@ -0,0 +1,223 @@
...
+  /**
+   * Generate the fingerprint data object for a script based on its current
+   * configuration, input files, and the fingerprints of its dependencies.
+   */
+  static async compute(
+    script: ScriptConfig,
+    dependencyFingerprints: Array<[ScriptReference, Fingerprint]>
+  ): Promise<Fingerprint> {
+    let allDependenciesAreCacheable = true;
+    const filteredDependencyStates: Array<
```
augustjk, line 115: Should this be renamed to `filteredDependencyFingerprints`?

src/execution/one-shot.ts
```diff
@@ -180,158 +202,97 @@ export class OneShotExecution extends BaseExecution<OneShotScriptConfig> {
...
+  /**
+   * Handle the outcome where the script was stale and we need to run it.
+   */
+  async #handleNeedsRun(fingerprint: Fingerprint): Promise<ExecutionResult> {
+    const shouldClean = await this.#shouldClean(fingerprint);
 
     // It's important that we delete the previous fingerprint and stdio replay
     // files before running the command or restoring from cache, because if
     // either fails mid-flight, we don't want to think that the previous
     // fingerprint is still valid.
```
augustjk, line 285: The "or restoring from cache" portion of this comment doesn't belong here now, but should the same note be added above the `await this.#prepareDataDir();` inside `#handleCacheHit()`?

src/fingerprint.ts
aomarks: Done thanks.

augustjk (approved):

rictic (approved):

PR #266, opened 2022-05-30
Refactor watcher to support analysis recovery and be more efficient

aomarks:
- Refactored the `Watcher` class to be easier to follow. Follows the (updated) state machine diagram more directly now.

- Analysis errors are no longer fatal. We collect all relevant `package.json` files during analysis even in the case of a failure, and then analyze again if any of them change.

- Analysis is now re-used across iterations as long as the `package.json` files didn't change, instead of re-analyzing on every iteration.

- Filesystem watchers are re-used across iterations as long as the script's input file config didn't change, instead of re-creating all watchers on every iteration.

Fixes https://github.com/google/wireit/issues/48
Supersedes https://github.com/google/wireit/pull/188

### DIFF
--- a/src/test/util/filesystem-test-rig.ts
+++ b/src/test/util/filesystem-test-rig.ts
@@ -83,6 +83,38 @@ export class FilesystemTestRig {
   }
 
   /**
+   * Like {@link write}, but first writes to a temporary file, and then renames
+   * the file to its final location.
+   */
+  async writeAtomic(file: string, content: unknown): Promise<void>;
+  async writeAtomic(files: {[filename: string]: unknown}): Promise<void>;
+  async writeAtomic(
+    fileOrFiles: string | {[filename: string]: unknown},
+    data?: string
+  ): Promise<void> {
+    this.assertState('running');
+    if (typeof fileOrFiles === 'string') {
+      const actual = pathlib.resolve(this.temp, fileOrFiles);
+      const temp = actual + '.tmp';
+      await this.write(temp, data);
+      await this.rename(temp, actual);
+    } else {
+      await Promise.all(
+        Object.entries(fileOrFiles).map(async ([relative, data]) =>
+          this.writeAtomic(relative, data)
+        )
+      );
+    }
+  }
+
+  /**
+   * Rename a file in the temporary filesystem.
+   */
+  async rename(oldPath: string, newPath: string): Promise<void> {
+    await fs.rename(this.resolve(oldPath), this.resolve(newPath));
+  }
+
+  /**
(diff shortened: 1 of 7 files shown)


src/test/util/filesystem-test-rig.ts
```diff
@@ -82,6 +82,38 @@ export class FilesystemTestRig {
     }
   }
 
+  /**
+   * Like {@link write}, but first writes to a temporary file, and then renames
+   * the file to its final location.
+   */
+  async writeAtomic(file: string, content: unknown): Promise<void>;
```
rictic, line 89: oh nice, yeah, that could totally be causing some of the race conditions I was seeing on my PR. maybe this should be the default for write?

src/test/watch.test.ts
```diff
@@ -498,31 +498,102 @@ test(
 );
 
 test(
-  'error from analysis is fatal (temporary)',
+  'recovers from analysis errors',
```
rictic, line 501: beautiful test

src/watcher.ts
```diff
@@ -28,245 +61,333 @@ import type {FailureMode} from './executor.js';
...
+        void oldWatcher.watcher.close();
+        this.#inputFileWatchers.delete(oldKey);
+      }
+    }
+  }
+
+  #onAbort(): void {
+    switch (this.#state) {
+      case 'watching':
+      case 'queued': {
```
rictic, line 324: 'queued' is like 'running', only we plan to start another run once it's finished yeah? In that case, I think we should treat it like 'running' rather than like 'watching' here.

as is, I think this could resolve #finished before the current execution is done

rictic (changes requested):
Great change! A few small suggestions and one substantive comment around how we treat an abort while queued

src/test/util/filesystem-test-rig.ts
aomarks: Yeah this really confused me for a while! I never hit it locally, or in CI on Linux or macOS, but on Windows CI I would get "invalid JSON" errors at weird times on maybe 1/4 of the runs. I think the filesystem is fast enough everywhere else that it didn't manifest.

I think we're unlikely to hit this kind of problem in any tests apart from the watch mode tests, because everywhere else we wait for the write to complete before we invoke Wireit.

I updated all the write calls in the watch mode tests to use writeAtomic, though, since similar problems could theoretically occur when modifying input files in watch mode too.

src/test/watch.test.ts
imhunterand: it's look a awesome test.

Issue #34, opened 2022-03-22
Pass-through extra arguments


aomarks:

jpzwarte:
So I have this:
```
"wireit": {
  "release": {
    "dependencies": [
      "./packages/foo:release",
      "./packages/bar:release"
    ]
  }
}

// foo & bar
"wireit": {
  "release": {
    "command": "npm publish"    
  }
}
```

I would like to be able to pass extra arguments on the cli, for example:
```
npm run release --tag beta
```

PR #272, opened 2022-06-02
Support extra arguments on command-line

aomarks:
You can now pass extra arguments to a script from the command-line like this:

```shell
npm run build -- -- --verbose
```

The two double-dashes are needed because the first one tells npm to forward arguments to wireit, and the second one tells wireit to forward arguments to the script.

It is also now an error to pass any unexpected argument to wireit (expected argument are currently just `watch` and `--`). Previously they were silently eaten.

It would be possible for us to relax the requirement for the second double-dash, and automatically pass through any arguments which we don't recognize as our own. However, that means every time we add a new argument, we'll need a breaking change. I think this is something we could relax in the future, but unless we get rid of arguments entirely, we'll still want `--` to mean "the following argument is for the script, not wireit" (e.g. if you wanted to pass a `watch` argument to the script). So I think this PR is a good starting point, with the option to relax later.

Fixes https://github.com/google/wireit/issues/34
cc @jpzwarte

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ npm run build -- -- --verbose
 
 The first double-dash tells `npm` to pass the remaining arguments to Wireit, and
 the second double-dash tells Wireit to pass the remaining arguments to the
-command.
+command. Only the `build` script will receive the `--verbose` argument.
 
 ## Input and output files
 
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@
   - [Vanilla scripts](#vanilla-scripts)
   - [Cross-package dependencies](#cross-package-dependencies)
 - [Parallelism](#parallelism)
+- [Extra arguments](#extra-arguments)
 - [Input and output files](#input-and-output-files)
 - [Incremental build](#incremental-build)
 - [Caching](#caching)
@@ -204,6 +205,18 @@ simultaneously, then only one instance will be allowed to run at a time, while
 the others wait their turn. This prevents coordination problems that can result
 in incorrect output files being produced. If `output` is set to an empty array,
 then this restriction is removed.
+
+## Extra arguments
+
+You can pass extra arguments to a Wireit script with _two_ double-dashes:
+
+```sh
+npm run build -- -- --verbose
+```
+
+The first double-dash tells `npm` to pass the remaining arguments to Wireit, and
+the second double-dash tells Wireit to pass the remaining arguments to the
+command.
 
 ## Input and output files
 
@@ -589,6 +602,7 @@ build](#incremental-build), and whether its output can be [restored from
 cache](#caching).
 
 - The `command` setting.
+- The [extra arguments](#extra-arguments) set on the command-line.
 - The `clean` setting.
 - The `output` glob patterns.
 - The SHA256 content hashes of all files matching `files`.
--- a/src/test/util/test-rig-command-interface.ts
+++ b/src/test/util/test-rig-command-interface.ts
@@ -72,46 +72,29 @@ export const MESSAGE_END_MARKER = '\x1e';
 /**
  * Sends and receives messages over an IPC data stream.
  */
-export class IpcClient<Incoming, Outgoing> {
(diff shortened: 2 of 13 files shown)


rictic:
Can we distinguish between `npm run foo watch` and `npm run foo -- watch`? If so, we could just support the former, and pass all arguments like the latter through to the underlying script

That would probably mean that we couldn't support `-k val` or `--key val` or `--key=val` style arguments but we might not need them

aomarks:
> Can we distinguish between npm run foo watch and npm run foo -- watch? If so, we could just support the former, and pass all arguments like the latter through to the underlying script

That would be nice, but unfortunately not. What we get on argv and env looks the same in those cases.

README.md
```diff
@@ -205,6 +206,18 @@ the others wait their turn. This prevents coordination problems that can result
...
+
+You can pass extra arguments to a Wireit script with _two_ double-dashes:
+
+```sh
+npm run build -- -- --verbose
+```
+
+The first double-dash tells `npm` to pass the remaining arguments to Wireit, and
+the second double-dash tells Wireit to pass the remaining arguments to the
+command.
```
rictic, line 219: Mention that --verbose is only passed to the "build" script

src/test/util/test-rig-command-interface.ts
```diff
@@ -39,3 +39,108 @@ export interface ExitMessage {
...
+/**
+ * Indicates the end of a JSON message on an IPC data stream. This is the
+ * "record separator" ASCII character.
+ */
+export const MESSAGE_END_MARKER = '\x1e';
+
+/**
+ * Sends and receives messages over an IPC data stream.
+ */
+export class IpcClient<Incoming, Outgoing> {
```
rictic, line 75: Nice

src/test/util/test-rig-command-interface.ts
```diff
@@ -39,3 +39,108 @@ export interface ExitMessage {
...
+   *
+   * Note that each data event could contain a partial message, or multiple
+   * messages. The special MESSAGE_END_MARKER character is used to detect the end
+   * of each complete JSON message in the stream.
+   */
+  #onData = (data: Buffer) => {
+    if (this.#closed.settled) {
+      throw new Error('Connection is closed');
+    }
+    for (const char of data.toString()) {
```
rictic, line 127: Oh that's really elegant, going character by character through the incoming data string

The iteration protocol might make this slightly slow, same with the use of `shift`, but plenty fast for test

src/test/util/test-rig-command.ts
```diff
@@ -140,6 +142,22 @@ export class WireitTestRigCommandInvocation {
...
   }
 
+  async environment(): Promise<Exclude<EnvironmentResponseMessage, 'type'>> {
+    this.#assertState('connected');
+    const client = new IpcClient<ChildToRigMessage, RigToChildMessage>(
+      this.#socket
+    );
+    client.send({type: 'environmentRequest'});
+    for await (const message of client.receive()) {
+      switch (message.type) {
```
rictic, line 152: Hm, this silently discards all other messages until an environmentResponse is received

I wonder if this class should extend IpcClient, override the receive method, and pass along everything but 'environmentResponse' which it handles by matching up with an environmentRequest

rictic (changes requested):

src/test/util/test-rig-command.ts
aomarks: Done, or something like it.

README.md

augustjk (approved):
🎉

rictic (approved):

aomarks:
I've discovered that yarn and pnpm have inconsistent behavior here, and I've also discovered that it is possible to recover `--foo` from `npm run --foo`, and to distinguish it from `npm run -- --foo`. So I think we can actually switch from `npm run watch` to `npm run --watch`, and then only a single `--` will be required for extra args. Should work in yarn and pnpm too. So I'll be sending another PR before releasing this feature.

PR #274, opened 2022-06-04
New --watch and extra args syntax

aomarks:
This PR is a follow up to https://github.com/google/wireit/pull/272 which changes the syntax for the watch mode flag and extra arguments.

Fixes https://github.com/google/wireit/issues/34
cc @jpzwarte

### Old syntax

```sh
npm run build watch -- -- --extra-args
```

### New syntax
```
npm run build --watch -- --extra-args
```

### Rationale

1. It turns out that npm allows us to capture the `--watch` argument from `npm run build --watch` by reading the `npm_config_watch` environment variable. (I did not realize this was possible when originally thinking about argument handling).

2. npm allows us to distinguish between `npm run build --watch` and `npm run build -- --watch`. However, it does *not* allow us to distinguish between `npm run build watch` and `npm run build -- watch`. This was the reason we needed the double `-- --` before, so now one is sufficient.

3. Only having to write one `--` seems much more natural, and exactly matches the built-in npm behavior. Writing `--watch` vs `watch` also seems more natural, since most other tools use the `--flag` style.

4. yarn and pnpm handle arguments differently to npm, and were not compatible with the changes I made in https://github.com/google/wireit/pull/272. Both yarn and pnpm actually allow us to reconstruct all of the command-line arguments, so while they each need special handling, we can match the behavior of this new syntax. They are also now tested.

### Other changes

1. Factored out a `cli-options` module because that code is now quite long, and for the next item...

2. Created a new set of `cli-options` tests that directly check the parsed options. This is more convenient and faster than the full integration tests, where we have to infer what options were parsed from the behavior.

3. Reverted the (unreleased) change from https://github.com/google/wireit/pull/272 which made unknown arguments an error. The problem with this is that Wireit arguments are now going to appear as `npm_config_<flag>` environment variables, and there are a bunch of `npm_config_*` variables that also get set that we would have to allow-list in order to decide which ones are unrecognized. Since those could change across npm versions or configurations, it seems risky to try and account for them all. Given that npm itself already silently ignores any `--flags` before a `--`, and we're trying to act like an extension to npm, I think this is fine.

4. Documented the support policy for npm/yarn/pnpm.


src/test/cli-options.test.ts
```diff
@@ -0,0 +1,249 @@
...
+      await assertOptions(rig, `${agent} start --watch -- --extra`, {
+        script: {
+          packageDir: rig.temp,
+          name: 'start',
+        },
+        extraArgs: ['--extra'],
+        watch: true,
+      });
+    })
+  );
```
rictic, line 246: I think for a couple versions we should detect someone doing using `watch` instead of `--watch` and give them a special error message, just because this change is going to catch people (and their existing scripts and workflows)

Something like:

```
As of wireit v0.6, use `--watch` instead of `watch`. This was added to support passing flags to scripts on the command line.

This will be an error for a couple of versions, and after that the `watch` argument will be passed to the underlying script.
```

augustjk: I second this.

aomarks: Done

src/cli-options.ts
```diff
@@ -0,0 +1,362 @@
...
+    return {
+      ok: false,
+      error: {
+        script,
+        type: 'failure',
+        reason: 'invalid-usage',
+        message:
+          `As of wireit v0.6, use "--watch" instead of "watch". ` +
+          `In an upcoming release, the "watch" argument will be passed to the script, ` +
+          `consistent with how npm usually behaves.`,
```
augustjk, line 192: Probably not a big deal for something meant to be temporary, but `watch` getting passed to script is only true for npm and gets swallowed as `unrecognized` for yarn and pnpm right?

augustjk (approved):

src/cli-options.ts
aomarks: Yep, for yarn and pnpm it will actually be a warning about unrecognized arguments, unless they put it on the right-hand-side of the `--`. Only npm has the weird behavior where some arguments get forwarded and some turn into environment variables.

But I'm ok with writing the docs around npm behavior to keep things simple, since that's our primary target. It should be fairly clear to the yarn/pnpm user when this ultimately happens, since they will see the warning.

PR #287, opened 2022-06-09
Allow `yarn run -T -B wireit`

### LINKED ISSUE #286: Support for Yarn Berry
I have a bunch of related projects.  Each has a structure like this:

```
packages/core
packages/preact
packages/...
```

where one package provides the framework-agnostic functionality, and sibling packages provide additional features for a particular framework.

Inspired by wireit, I'm exploring merging them into a monorepo with this structure:

```
project-a/packages/core
project-a/packages/preact
project-a/packages/...
project-b/packages/core
project-c/packages/preact
project-c/packages/...
```

To do that, I need to upgrade to Yarn Berry. In Berry, `"build": "wireit"` will only resolve in the current package.json.  To use `devDependencies` shared with sibling packages, [one must use `yarn run -T wireit`](https://yarnpkg.com/cli/run).  This causes the "you must use `wireit` naked" error to be thrown.

Would you be amenable to a PR that amends this line:
https://github.com/google/wireit/blob/48839ed5da4de5e0df59f6e9931a61d6efb9503b/src/analyzer.ts#L338

to instead check
```ts
[`wireit`, `yarn run -T wireit`].includes(scriptCommand.value)
```

This will provide the same safety (that wireit isn't being chained into a larger command), but will allow a single wireit installation to be shared across sibling packages in modern versions of yarn.

appsforartists:
yarn@2+ (Berry) distinguishes between local scripts (`yarn run script-name`) and scripts in the workspace root (`yarn run -T script-name`).  This PR relaxes the script assertion to allow wireit to be invoked with `yarn run -T wireit`, enabling projects managed by Berry to use a shared instance of yarn at the top level.

Note: I do not know how to properly write tests for this.

Fixes #286

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -335,7 +335,11 @@ export class Analyzer {
     const scriptCommand = syntaxInfo.scriptNode;
     const wireitConfig = syntaxInfo.wireitConfigNode;
 
-    if (wireitConfig !== undefined && scriptCommand.value !== 'wireit') {
+    if (
+      wireitConfig !== undefined &&
+      scriptCommand.value !== 'wireit' &&
+      scriptCommand.value !== 'yarn run -T -B wireit'
+    ) {
       const configName = wireitConfig.name;
       placeholder.failures.push({
         type: 'failure',
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,15 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - You can now pass arbitrary extra arguments to a script by setting them after
   two double-dashes, e.g. `npm run build -- -- --verbose`.
 
+- If you're using Yarn Berry, you can invoke the shared instance of wireit at
+  the root of your workspace from any package's `scripts` entry:
+
+  ```json
+  "scripts": {
+    "build": "yarn run -T -B wireit"
+  },
+  ```
+
 ### Changed
 
 - [**Breaking**] Setting unrecognized command-line arguments is now an error.


src/analyzer.ts
```diff
@@ -335,7 +335,10 @@ export class Analyzer {
     const scriptCommand = syntaxInfo.scriptNode;
     const wireitConfig = syntaxInfo.wireitConfigNode;
 
-    if (wireitConfig !== undefined && scriptCommand.value !== 'wireit') {
+    if (
+      wireitConfig !== undefined &&
+      ![`wireit`, `yarn run -T wireit`].includes(scriptCommand.value)
```
aomarks, line 340: Seems like `yarn run -T -B wireit` would be good (though mind verifying that works?), since that enforces it is a binary, and not a local script called `wireit`?

aomarks, line 340: Would prefer `scriptCommand.value !== 'wireit' && scriptCommand.value !== 'yarn run -T -B wireit'`, to avoid the extra array allocation.

aomarks:
> Note: I do not know how to properly write tests for this.

We'd need a way to have both versions of yarn installed at the same time. I suppose we could also have a separate workflow that switches out the version. Hmm. I will think about it.

Have you tested this manually by linking to another package?

aomarks (changes requested):
Thanks for the PR!

aomarks (changes requested):
Also if you wouldn't mind, we track all changes in the `CHANGELOG.md` file. This would go under `Added` I think.

src/analyzer.ts
appsforartists: It's your project, so I'm happy to make whatever changes you prefer.

Does the array allocation matter?  Seems like this runs once at the beginning, where the impact of an array is negligible compared to the benefits to {read/maintain}ability.

appsforartists: (with prettier, it ends up like this:
```ts
    if (
      wireitConfig !== undefined &&
      scriptCommand.value !== 'wireit' &&
      scriptCommand.value !== 'yarn run -T -B wireit'
    ) {
```
so the array isn't significantly more readable)

appsforartists:
Yeah, I've tested it manually.  `-T -B` works.

Thanks Al!

appsforartists:
The Ubuntu failure is GitHub Actions being flaky.

aomarks (approved):

PR #290, opened 2022-06-10
Temporary CI fix for npm bug

aomarks:
Context: https://github.com/npm/cli/issues/4980, https://github.com/lit/lit/pull/3011

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,14 @@ jobs:
       # We support Node Latest, Active LTS, and the most recent Maintenance LTS.
       # See https://nodejs.org/en/about/releases/ for Node release schedule.
       matrix:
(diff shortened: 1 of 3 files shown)


augustjk (approved):

augustjk:
Oh, oops, it's actually not just limited to 16. The default latest patch versions for 14 and 18 also are shipping with the problematic npm.

And the repo setting for required workflow seems like it needs updating.

Perhaps a better alternative might be to change the affected assertions to `match` rather than `equal`?

augustjk:
Or doing the workaround mentioned here https://github.com/actions/virtual-environments/issues/5716#issuecomment-1152038439 as a windows-only workflow step might be worth trying?

aomarks:
Fixed it by just removing the culprit string from `stderr` as we report it from the test rig.

Issue #294, opened 2022-06-10
Build fails in Node 14 workflow


augustjk:
Some node modules are not installed when running `npm ci` in node 14 which causes errors during the typescript build. Specifically `vscode-languageserver`, `vscode-languageserver-textdocument`, and `vscode-languageclient` are needed for the build to be successful.

See https://github.com/google/wireit/runs/6838811493?check_suite_focus=true

This is probably due to npm v6 not supporting workspaces and those dependencies are only listed in `vscode-extension/package.json`.

Potential solutions:
1. Reorganize the files so that `language-server.ts` and `ide.ts` and related tests could be part of the `vscode-extension` package (whose tests are not run in node 14)
2. Split them into a separate package like `language-server`
3. Configure the root project to have separate typescript builds so the `test:headless` can depend on that and run.

rictic:
I think we could just make them dev dependencies in the main wireit package.json

PR #295, opened 2022-06-10
Make tests workflow run with correct node versions

augustjk:
Previously all the workflows were running in node 16 as the matrix key was not correct. This fixes the workflow yaml to actually use the correct versions. Also added a step to check the versions of node and npm within the workflow.

However, this did surface an issue for node 14. https://github.com/google/wireit/issues/294
I've commented out 14 from the matrix for now but if this is to be merged as is, the repo workflow requirement would need to be loosened to not require 14.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,12 @@ jobs:
       # We support Node Latest, Active LTS, and the most recent Maintenance LTS.
       # See https://nodejs.org/en/about/releases/ for Node release schedule.
       matrix:
-        version: [14, 16, 18]
+        version: [
+            # Skip 14 for now. See https://github.com/google/wireit/issues/294
+            # 14,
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -8,7 +8,12 @@ jobs:
       # We support Node Latest, Active LTS, and the most recent Maintenance LTS.
       # See https://nodejs.org/en/about/releases/ for Node release schedule.
       matrix:
-        version: [14, 16, 18]
+        version: [
```
aomarks, line 11: I kinda prefer "node" as the key, because "version" is more ambiguous ("os" is also a kind of version)

augustjk: Yes, that is totally true. Done!

aomarks (approved):

Issue #292, opened 2022-06-10
Output files should be validated before assuming a script is fresh


aomarks:
### Background

Currently if the `.wireit/<script>/fingerprint` file matches the current fingerprint for a script, then we assume that script is fresh, and skip it.

This works great, unless another process modifies the output files independently of Wireit.

The following issues are all examples of ways users have inadvertently gotten their script into this state:

https://github.com/google/wireit/issues/150
https://github.com/google/wireit/issues/262
https://github.com/google/wireit/issues/280

### Proposal

Changes:

1. After running a script, list all files matching the `output` globs, and save a record of their metadata (ctime, mtime, size, inode number) into a file named `.wireit/<script>/manifest`.

2. When preparing to run a script, after noticing that the current fingerprint matches the `.wireit/<script>/fingerprint` file, re-compute the manifest (in the same way as we did in [1]), and compare it to the `.wireit/<script>/manifest` file. If they match, consider the script fresh, and skip. If they don't match, consider the script stale, and act the same way as we would if the fingerprint didn't match.

Issue #280, opened 2022-06-07
Problems when switching branches


sorvell:
I generally have problems with wireit when switching between github branches. If a file that the `.wireit` cache expects to exist on disk is deleted (but the `.wireit` cache is not, and it's typically in `.gitignore`), then builds will fail due to `wireit` thinking the build is fresh when the build output is not there. 

It would be great if this were either somehow handled automatically or there was an easy way to address the issue (e.g. better than having to nuke all the `.wireit` files).

aomarks:
https://github.com/google/wireit/issues/292 Should resolve this issue, so will track there.

aomarks:
Wireit 0.6.1 was just released which includes output file verification, which should resolve this issue.

Issue #262, opened 2022-05-27
command gets re-executed even with no change in input files


SanderElias:
This might be me misunderstanding the nuances in the config. However, the command gets called every time I run `npm run buildDeps`

```json
     "buildServer": {
      "command": "tsc --pretty --project server/tsconfig.server.json",
      "clean": "if-file-deleted",
      "files": [  "./server/**/*.mts", "./server/**/*.html",  "./server/tsconfig.server.json"  ],
      "output": [  "./dist/server" ]
    },
    "sharedSock": {
      "command": "ng build --configuration production sharedSock",
      "clean": true,
      "files": [ "./projects/shared-sock/src/lib/**/*.ts"  ],
      "output": [ "./dist/shared-sock"  ]
    },
    "buildDeps": {
      "dependencies": [  "buildServer", "sharedSock"  ]
    },
```

I tried quite a number of things, but the command for the `"sharedSock"` dependency gets executed every time. 
It might be me misunderstanding the setup, but it should skip the command if there is no change in the `files` of the `"sharedSock"` right?
I looked at #238 which pointed me to #245, but I seem to already have taken those steps?

aomarks:
Based on the config shown, it seems like you should be getting fresh skips/caching. Is there anything more to know about how you are running the command? If you directly run `npm run sharedSock` twice in a row, does it execute twice?

SanderElias:
@aomarks Yes, it does execute twice. 
In watch mode, it will execute on every change, including changes in the server

aomarks:
(Sorry, I incorrectly referenced this issue which auto-closed it)

milesrichardson:
I had a similar issue and was able to solve it by adding some exclude patterns to `output`. In my case I was generating `.generated.ts` files for any `.ts` file that includes some GQL snippet that `graphql-codegen` wants to use. The problem was that on the second run, `wireit` would report the output as "fresh", but copy files from the cache, and in doing so, delete the `.generated.ts` files. 

The solution was to add two entries to `output`: `!build`, which was a simple oversight by me, but also weirdly `!.wireit`, which seemed like a bug. FWIW, this was in a child package, not the root.

[...]

Issue #150, opened 2022-04-28
Output files are not restored from cache


mihar-22:
Thanks heaps for open-sourcing this having some fun already, it's awesome. I'm not sure if I'm doing something silly but it seems when I run the `build` command as shown below the files are not restored from cache. Is this expected behaviour because my assumption was it would restore the output files?

![image](https://user-images.githubusercontent.com/14304599/165668463-bb5400c2-26fa-4258-94da-bb9bae1a3353.png)

aomarks:
Hmm, that configuration definitely looks like it should work with caching. Caching to local disk should be enabled if you have both `files` and `output`.

Can you try `ls .wireit/*/cache`? That's where local output caches are saved, so whether there is something there or not would be a clue.

(One side note is that the `packageLocks` currently need to be in _every_ script, but I see you already noticed https://github.com/google/wireit/issues/66).

mihar-22:
I can confirm files are being cached but they're just not being output:

![image](https://user-images.githubusercontent.com/14304599/165675232-e3dd0f88-35c0-445b-82aa-700a047fef85.png)

mihar-22:
To be clear what I mean by restoring is when I manually or through a script delete the output (i.e., `types/` dir), it isn't put back on a subsequent wireit run.

aomarks:
Oh I see.

Wireit does not use the output files of a script to determine whether it should restore from cache or not -- only the input files and some other fields determine that (full list at https://github.com/google/wireit#cache-key) .

So deleting or changing output files of a script independently of the script can confuse things.

If you need to recover from a situation like this, type `rm -rf .wireit/*/state`. This will cause Wireit to either rerun the script or restore them from cache.

mihar-22:
Makes sense, I guess I was coming from Turborepo and I was accustomed to the output files being put back. I guess what I'm asking for is more valuable when remote caching is involved (which I would love to see). Thanks for the swift response.

PR #300, opened 2022-06-14
Verify output files before treating a script as fresh

aomarks:
### Background

Before this PR, if the `.wireit/<script>/fingerprint` file matches the current fingerprint for a script, then we assume that script is fresh, and skip it.

This works great, unless another process modifies the output files independently of Wireit, in which case the script will be skipped even though it actually needed to run.

### This PR

Now, after running a script, we now list all files matching the `output` globs, and save a record of their metadata (name, mtime, size) into a file named `.wireit/<script>/manifest`.

Then when preparing to run a script, after noticing that the current fingerprint matches the `.wireit/<script>/fingerprint` file, we re-compute the manifest (in the same way as we did above), and compare it to the `.wireit/<script>/manifest` file. If they match, consider the script fresh, and skip. If they don't match, consider the script stale, log an info message, and act the same way as we would if the fingerprint didn't match.

Fixes https://github.com/google/wireit/issues/292
Fixes https://github.com/google/wireit/issues/280
Fixes https://github.com/google/wireit/issues/262
Fixes https://github.com/google/wireit/issues/150

### Other changes

- Previously, if `output` was undefined, we could still treat a script as fresh, because we assumed the output wouldn't have changed if the input files didn't change. We no longer assume this, so now it's necessary to specify `output` to get freshness skipping (can be empty array).

- Glob results are now always returned as absolute paths. This made it simpler to share the cached output glob results between manifest generation and cleaning.

- Fixed a test rig bug where `WIREIT_FAILURES` would get propagated to the test cases.

### DIFF
--- a/src/caching/github-actions-cache.ts
+++ b/src/caching/github-actions-cache.ts
@@ -18,7 +18,7 @@ import type {Cache, CacheHit} from './cache.js';
 import type {ScriptReference} from '../script.js';
 import type {Fingerprint} from '../fingerprint.js';
 import type {Logger} from '../logging/logger.js';
-import type {RelativeEntry} from '../util/glob.js';
+import type {AbsoluteEntry} from '../util/glob.js';
 import type {Result} from '../error.js';
 
 /**
@@ -147,18 +147,18 @@ export class GitHubActionsCache implements Cache {
   async set(
     script: ScriptReference,
     fingerprint: Fingerprint,
-    relativeFiles: RelativeEntry[]
+    absFiles: AbsoluteEntry[]
   ): Promise<boolean> {
     if (this.#hitRateLimit) {
       return false;
     }
 
-    const absFiles = relativeFiles.map((rel) =>
-      pathlib.join(script.packageDir, rel.path)
-    );
     const tempDir = await makeTempDir(script);
     try {
-      const tarballPath = await this.#makeTarball([...absFiles], tempDir);
+      const tarballPath = await this.#makeTarball(
+        [...absFiles.map((file) => file.path)],
+        tempDir
+      );
       return await this.#reserveUploadAndCommitTarball(
         script,
         fingerprint,
--- a/src/execution/one-shot.ts
+++ b/src/execution/one-shot.ts
@@ -27,6 +27,7 @@ import type {Logger} from '../logging/logger.js';
 import type {WriteStream} from 'fs';
 import type {Cache, CacheHit} from '../caching/cache.js';
 import type {StartCancelled} from '../event.js';
+import type {AbsoluteEntry} from '../util/glob.js';
 
 type OneShotExecutionState = 'before-running' | 'running' | 'after-running';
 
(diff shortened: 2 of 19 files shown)


src/util/manifest.ts
```diff
@@ -0,0 +1,61 @@
...
+  m: number | undefined;
+  /** Size in bytes. `undefined` for directories. */
+  s: number | undefined;
+}
+
+/**
+ * A JSON-serialized manifest of files.
+ */
+export type FileManifestString = string & {
+  __OutputManifestStringBrand__: never;
```
augustjk, line 34: Super nit. Should it be `__FileManifestStringBrand__` to match the type name?

src/execution/one-shot.ts
```diff
@@ -458,45 +502,86 @@ export class OneShotExecution extends BaseExecution<OneShotScriptConfig> {
...
+
+  /**
+   * Glob the output files for this script and cache them, but throw unless the
+   * script has not yet started running or been restored from cache.
+   */
+  #globOutputFilesBeforeRunning(): Promise<
+    Result<AbsoluteEntry[] | undefined>
+  > {
+    this.#ensureState('before-running');
+    return (this.#cachedOutputFilesBeforeRunning ??= this.#globOutputFiles());
```
augustjk, line 550: Neat!

src/test/freshness.test.ts
```diff
@@ -1749,4 +1838,107 @@ test(
   })
 );
 
+test(
+  'script is not fresh if output file is modified externally',
```
augustjk, line 1842: Love this test!

augustjk (approved):
Can't wait for this to be released!

src/execution/one-shot.ts
```diff
@@ -458,45 +502,86 @@ export class OneShotExecution extends BaseExecution<OneShotScriptConfig> {
...
+    Result<AbsoluteEntry[] | undefined>
+  >;
+
+  /**
+   * Glob the output files for this script and cache them, but throw unless the
+   * script has finished running or been restored from cache.
+   */
+  #globOutputFilesAfterRunning(): Promise<Result<AbsoluteEntry[] | undefined>> {
+    this.#ensureState('after-running');
+    return (this.#cachedOutputFilesAfterRunning ??= this.#globOutputFiles());
```
rictic, line 562: elegant!

rictic (approved):

AndrewJakubowicz (commented):
Awesome PR!

src/caching/github-actions-cache.ts
```diff
@@ -147,18 +147,18 @@ export class GitHubActionsCache implements Cache {
...
     }
 
-    const absFiles = relativeFiles.map((rel) =>
-      pathlib.join(script.packageDir, rel.path)
-    );
     const tempDir = await makeTempDir(script);
     try {
-      const tarballPath = await this.#makeTarball([...absFiles], tempDir);
+      const tarballPath = await this.#makeTarball(
+        [...absFiles.map((file) => file.path)],
```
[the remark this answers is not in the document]
aomarks: Good point, done in https://github.com/google/wireit/pull/288

PR #363, opened 2022-07-26
Reduce default worker count to 2x logical CPU core count

augustjk:
Doing `4 * os.cpus().length` seems a little too taxing on the system. When many tests are run in parallel, a test can take significantly longer causing it to hit timeouts that it otherwise would not have hit when run standalone, requiring increasing the timeout as seen here https://github.com/lit/lit/pull/3182. Reducing to 2x seems more reasonable.

Also updated the docs to say the default is a multiple of "logical CPU cores", which is the count of the number threads the OS sees the system can process, instead of just "CPUs" or "CPU cores", in line with Node's [doc](https://nodejs.org/api/os.html#oscpus) on what `os.cpus()` returns.


aomarks (approved):
I agree being a little more conservative is a good idea given our experience with Lit monorepo CI.

PR #437, opened 2022-09-25
Add single task debounce to watcher

aomarks:
File change events are now debounced using a `setTimeout` with 0ms. In the future we could maybe make this a user-controlled parameter, but the immediate purpose is to solve a bug I noticed while setting up the webcomponents.org dev configuration.

The bug was that if multiple scripts have the same input files, then because in watch mode we register a filesystem watcher for *each of those scripts* *, when a common file changed, we'd start a run from the first event, and then immediately queue up a second run from the second event. Now we wait a task before running, so the two events get batched.

For a build with good caching configuration, this would usually cause a no-op second build -- but that's quite annoying because of the console spam. For a build without caching, this would cause a fully redundant build.

**(We could potentially register fewer watchers by noticing when input files are the same across scripts, but note that different globs can match the same file, so it's often not possible to do this, and the bug would still exist.)*

### DIFF
--- a/src/test/watch.test.ts
+++ b/src/test/watch.test.ts
@@ -755,4 +755,67 @@ test(
   })
 );
 
+test(
+  'debounces when two scripts are watching the same file',
+  timeout(async ({rig}) => {
+    const cmdA = await rig.newCommand();
+    const cmdB = await rig.newCommand();
+    await rig.writeAtomic({
+      'package.json': {
+        scripts: {
+          a: 'wireit',
+          b: 'wireit',
+        },
+        wireit: {
+          a: {
+            command: cmdA.command,
+            dependencies: ['b'],
+            files: ['input.txt'],
+            // Note it's important for this test that we don't have output set,
+            // because otherwise the potential third run would be restored from
+            // cache, and we wouldn't detect it anyway.
+          },
+          b: {
+            command: cmdB.command,
+            files: ['input.txt'],
+          },
+        },
+      },
+      'input.txt': 'v0',
+    });
+
+    const exec = rig.exec('npm run a --watch');
+
+    // Initial run.
+    {
+      (await cmdB.nextInvocation()).exit(0);
+      (await cmdA.nextInvocation()).exit(0);
+    }
+
+    // Wait until wireit is in the "watching" state, otherwise the double file
+    // change events would occur in the "running" state, which wouldn't trigger
+    // the double runs.
+    await exec.waitForLog(/Watching for file changes/);
+
+    // Changing an input file should cause one more run.
+    {
+      await rig.writeAtomic({
+        'input.txt': 'v1',
+      });
+      (await cmdB.nextInvocation()).exit(0);
+      (await cmdA.nextInvocation()).exit(0);
+    }
+
+    await exec.waitForLog(/Watching for file changes/);
+
+    // Wait a moment to ensure a third run doesn't occur.
+    await new Promise((resolve) => setTimeout(resolve, 100));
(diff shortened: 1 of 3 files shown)


src/test/watch.test.ts
```diff
@@ -755,4 +755,67 @@ test(
...
+        'input.txt': 'v1',
+      });
+      (await cmdB.nextInvocation()).exit(0);
+      (await cmdA.nextInvocation()).exit(0);
+    }
+
+    await exec.waitForLog(/Watching for file changes/);
+
+    // Wait a moment to ensure a third run doesn't occur.
+    await new Promise((resolve) => setTimeout(resolve, 100));
```
AndrewJakubowicz, line 812: Is this brittle (if the number is changed in the future)? Is there any public state to assert (e.g. if the state is in an unexpected state - or is that testing private state?).

AndrewJakubowicz (approved):
Excellent!

src/watcher.ts
```diff
@@ -139,10 +162,41 @@ export class Watcher {
     this.#abort = abort;
   }
 
-  #startRun(): void {
+  #startDebounce(): void {
```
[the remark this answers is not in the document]
aomarks: Sounds reasonable. I agree it should never be in that state.

src/test/watch.test.ts
aomarks: If we changed the default from 0 to something close to or above 100, then it would be brittle yeah. But I think it's unlikely we'd have a default that high. Not sure I can think of any other public state we could probe.

aomarks (commented):
Thanks for review!

src/watcher.ts
```diff
@@ -342,6 +420,9 @@ export class Watcher {
       case 'aborted': {
         return;
       }
+      case 'initial': {
```
rictic, line 426: why is it unexpected to abort before starting the first run?

rictic (approved):

src/watcher.ts
aomarks: Because we immediately transition to `running`. But I don't want the initial state to be `running` because in every other situation calling `startRun` while the state is already `running` indicates something bad.

src/watcher.ts
```diff
@@ -54,6 +64,18 @@ interface FileWatcher {
   watcher: chokidar.FSWatcher;
 }
 
+/**
+ * The minimum time that must elapse after the last file change was detected
+ * before we begin a new run. Also the minimum time between successive runs.
+ *
+ * Note even 0 is a useful value here, because that defers new runs to the next
+ * JS task. This is important because if multiple scripts are watching the same
+ * file that changed, we get a file watcher event for each of them. Without
```
justinfagnani, line 73: Is there a way to see the queue of FS events, consume all pending events synchronously similar to takeRecords in MutationObservers, or see if there are any pending events at all similar to isInputPending()? I'm wondering if in cases of multiple file changes you still get event 1, event 2, run (consume event 1), run (consume event 2).

justinfagnani: I was wondering how enqueued events were handled, and the queues are run through per event loop turn: https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/#phases-overview

aomarks: Cool. Though just for clarity, in the case of Wireit, we have a separate chokidar watcher for each script. So even if a chokidar watcher is batching all changes per JS task, we will still receive a separate event for each script. Hence it's still helpful for us to have our own debounce.

PR #444, opened 2022-10-02
Various prep work for services

aomarks:
Some misc stuff I've factored out while working on [services](https://github.com/google/wireit/issues/33).

**I recommend viewing commits one-by-one**

- Minor stuff
  - Use `~` for TypeScript dependency, since it doesn't follow semver.
  - Bump dependencies.
  - Use new `@types/selfsigned` package instead of locally defined types.
  - Un-nest some `Omits`.
  - Close our test command connections before killing their associated processes. The reverse order was causing tests to hang in some cases.
- Names
  - Renamed `cacheable` property to `fullyTracked`. I think this name is more clear, because while it's used to determine if something can be cached, it also determines whether something can be fresh. Plus services will be able to be fully tracked, but are never cached. What it really means is "we know about all of the inputs and outputs of this script".
  - Renamed `OneShot` to `Standard`. `OneShot` was just weird. We need a distinction between service and non-service scripts, but I don't think `OneShot` did that well.
  - Renamed `NoOp` to `NoCommand`. More precise, since it specifically means "a script with no command", and we don't use "op" elsewhere.
  - Renamed `script.ts` to `config.ts`. I keep expecting it to be called `config`, since it holds the config types.
- Output
  - Replace native `#private` fields with `private` keyword, and disable `useDefineForClassFields`. Since we still need to support Node 14, we can't emit native class fields. This means we currently emit very unreadable and inflated code to match class field semantics, which makes it much harder to read the code we distribute, which makes debugging much harder. We can go back to native class fields on 2023-04-30, which is the Node 14 end-of-life.


justinfagnani (approved):
Thanks for the easy-to-review commits!

PR #449, opened 2022-10-06
Remove test-rig tempfix for Windows

augustjk:
Remove the temp fix introduced in #290.
The latest versions of npm shipping in the Github actions runners no longer cause the original issue.


aomarks (approved):
Nice! How did you remember that :P

PR #451, opened 2022-10-06
Remove Alpha warning

aomarks:
I don't think we need it anymore. There are features to add, but the features we have are quite stable and proved out now.


augustjk (approved):
Yay! I agree it's working quite well.

rictic (approved):

PR #460, opened 2022-10-14
Services: Documentation and config

aomarks:
First of a series of PRs to implement [services](https://github.com/google/wireit/issues/33).

This just documents the feature and adds the config field.

Note this PR is merging into the `service` branch, where I'll be staging this feature until everything is ready (in case we need to release something in the meantime).

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -45,7 +45,7 @@
             "type": "array"
           },
           "service": {
-            "markdownDescription": "If true, treat this script as a long-running process.\nServices are automatically brought up and down as they depended upon by other scripts. If invoked directly, services continue running until Wireit is killed with Ctrl-C.\nFor more info, see: https://github.com/google/wireit#services",
+            "markdownDescription": "If true, treat this script as a long-running process.\nServices are automatically brought up and down as they are depended upon by other scripts. If invoked directly, services continue running until Wireit is killed with Ctrl-C.\nFor more info, see: https://github.com/google/wireit#services",
             "type": "boolean"
           }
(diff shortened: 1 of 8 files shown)


schema.json
```diff
@@ -43,6 +43,10 @@
               "type": "string"
             },
             "type": "array"
+          },
+          "service": {
+            "markdownDescription": "If true, treat this script as a long-running process.\nServices are automatically brought up and down as they depended upon by other scripts. If invoked directly, services continue running until Wireit is killed with Ctrl-C.\nFor more info, see: https://github.com/google/wireit#services",
```
rictic, line 48: as they depended upon -> as they are depended upon

rictic (approved):
Nice design

I can imagine potential improvements, like a way for wireit to communicate with a running service process (to indicate things like when the service is ready, etc), but those would be optional if added, and they might not be worth the complexity

schema.json
aomarks: Done

aomarks:
> I can imagine potential improvements, like a way for wireit to communicate with a running service process (to indicate things like when the service is ready, etc), but those would be optional if added, and they might not be worth the complexity

Yeah -- I think we probably should include a `waitForLog` or similar feature to check for a string or pattern on stdout/stderr, since that's pretty simple and will cover a lot of use-cases where you need to wait for a server to be fully started after it's spawned. E.g. `"waitForLog": "server listening"`

justinfagnani (approved):

PR #461, opened 2022-10-15
Services: Config types, events, stub execution class, and started promise

aomarks:
1. Added a `started` promise property to `ScriptChildProcess`. We'll be re-using this class for services, and we'll need to know when a child process has spawned. We didn't need that exposed before, because all that mattered was the exit code.

2. Added events and logging specific to services.

3. Refactored the config types so that classes can hold copies of just a subset of the fields of a config, instead of the full object. In watch mode, service child processes will get handed-off _across iterations_, and potentially _across different build graphs_, so I wanted to make sure we don't have a memory leak relating to keeping around references to old build graphs (configs hold references to their dependencies, so potentially that keeps the full build graph live).

4. Added a `ServiceScriptConfig` which is guaranteed to have `service:false`, and a stub `ServiceScriptExecution` which currently can produce a fingerprint, but doesn't yet actually start/stop the service command.

More incremental work towards [services](https://github.com/google/wireit/issues/33).

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -1082,19 +1082,44 @@ export class Analyzer {
       };
       return {ok: false, error: this._markAsInvalid(config, failure)};
     }
-    {
-      const validConfig: ScriptConfig = {
+
+    let validConfig: ScriptConfig;
+    if (config.service) {
+      // We should already have created an invalid script at this point, so we
+      // should never get here. We throw here to convince TypeScript that this
+      // is guaranteed.
+      if (config.command === undefined) {
+        throw new Error(
+          'Internal error: Supposedly valid service did not have command'
+        );
+      }
+      validConfig = {
         ...config,
-        extraArgs: undefined,
         state: 'valid',
+        extraArgs: undefined,
         dependencies: config.dependencies as Array<Dependency<ScriptConfig>>,
+        // Unfortunately TypeScript doesn't narrow the ...config spread, so we
+        // have to assign explicitly.
+        command: config.command,
       };
-      // We want to keep the original reference, but get type checking that
-      // the only difference between a ScriptConfig and a
-      // LocallyValidScriptConfig is that the state is 'valid' and the
-      // dependencies are also valid, which we confirmed above.
-      Object.assign(config, validConfig);
+    } else {
+      validConfig = {
+        ...config,
+        state: 'valid',
+        extraArgs: undefined,
+        dependencies: config.dependencies as Array<Dependency<ScriptConfig>>,
+        // Unfortunately TypeScript doesn't narrow the ...config spread, so we
+        // have to assign explicitly.
+        service: config.service,
+      };
     }
+
+    // We want to keep the original reference, but get type checking that
+    // the only difference between a ScriptConfig and a
+    // LocallyValidScriptConfig is that the state is 'valid' and the
+    // dependencies are also valid, which we confirmed above.
+    Object.assign(config, validConfig);
+
     return {ok: true, value: config as unknown as ScriptConfig};
   }
 
--- a/src/script-child-process.ts
+++ b/src/script-child-process.ts
@@ -13,7 +13,7 @@ import {
 import {Deferred} from './util/deferred.js';
 
 import type {Result} from './error.js';
-import type {ScriptConfig} from './config.js';
+import type {ScriptReferenceWithCommand} from './config.js';
 import type {ChildProcessWithoutNullStreams} from 'child_process';
 import type {ExitNonZero, ExitSignal, SpawnError, Killed} from './event.js';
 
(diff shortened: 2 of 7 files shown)


src/script-child-process.ts
```diff
@@ -44,27 +45,33 @@ export type ScriptChildProcessState =
...
+   */
+  get started(): Promise<Result<void, SpawnError>> {
+    return this._started.promise;
+  }
+
   /**
    * Resolves when this child process ends.
    */
-  readonly completed: Promise<
+  get completed(): Promise<
```
rictic, line 70: can this just be:

```typescript
readonly completed = this._completed.promise;
```

rictic (approved):

src/script-child-process.ts
justinfagnani: If there's even going to be a new run of the child process with new Deferreds for started and completion, then the getters are safer.

src/script-child-process.ts
```diff
@@ -113,94 +127,96 @@ export class ScriptChildProcess {
...
+          const exception = new Error(
+            `Internal error: Expected ScriptChildProcessState ` +
+              `to be "started" or "killing" but was "${this._state}"`
+          );
+          this._started.reject(exception);
+          this._completed.reject(exception);
+          break;
+        }
+        default: {
+          const never: never = this._state;
```
justinfagnani, line 156: Can this be folded into the previous case?

src/script-child-process.ts
```diff
@@ -113,94 +127,96 @@ export class ScriptChildProcess {
...
+          const exception = new Error(
```
justinfagnani, line 157: If you want the Error to have a stack trace, then you can throw it inside a try/catch and call reject() after it's caught.

src/script-child-process.ts
```diff
@@ -113,94 +127,96 @@ export class ScriptChildProcess {
...
+          message: error.message,
+        },
+      } as const;
+      this._started.resolve(result);
+      this._completed.resolve(result);
+      this._state = 'stopped';
+    });
+
+    this._child.on('close', (status, signal) => {
+      if (this._state === 'killing') {
```
justinfagnani, line 184: Do you need to ensure or assert that `_started` is settled in any of these cases?

src/analyzer.ts
```diff
@@ -1082,19 +1082,44 @@ export class Analyzer {
...
+      };
+    } else {
+      validConfig = {
+        ...config,
         state: 'valid',
+        extraArgs: undefined,
         dependencies: config.dependencies as Array<Dependency<ScriptConfig>>,
+        // Unfortunately TypeScript doesn't narrow the ...config spread, so we
+        // have to assign explicitly.
+        service: config.service,
```
justinfagnani, line 1113: Can you cast the config: `...(config as ServiceScriptConfig)`?

src/analyzer.ts
```diff
@@ -1082,19 +1082,44 @@ export class Analyzer {
...
+        // Unfortunately TypeScript doesn't narrow the ...config spread, so we
+        // have to assign explicitly.
+        command: config.command,
+      };
+    } else {
+      validConfig = {
+        ...config,
         state: 'valid',
+        extraArgs: undefined,
         dependencies: config.dependencies as Array<Dependency<ScriptConfig>>,
```
justinfagnani, line 1110: Why was this cast needed?

justinfagnani (approved):

src/script-child-process.ts
aomarks: Good point, done!

> If there's even going to be a new run of the child process with new Deferreds for started and completion, then the getters are safer.

The child process will only be re-used if it doesn't need to be restarted. Otherwise there will be a new `ScriptChildProcess` instance. So this is safe.

src/script-child-process.ts
aomarks: No, it's a different error, and it has a `never` assignment to prove to TypeScript that it can't happen. Or did I misunderstand somethning?

src/script-child-process.ts
aomarks: Cool trick. In this case the error string is unique, so I don't think a trace trace really adds much value.

src/script-child-process.ts
aomarks: I don't think so. `close` should only happen after `spawn` or `error`.

src/analyzer.ts
aomarks: Yes, but then we'd be bypassing the type checker all together. Here I'm trying to prove to the type checker (and hence ourselves) that the 3 different configs can be discriminated.

src/analyzer.ts
aomarks: Because that property requires `validated` dependencies, but the type checker still believes they could be `unvalidated`, because that's checked earlier in a way that doesn't cause the type checker to narrow.

PR #466, opened 2022-10-19
Services: Add service-related analysis properties

aomarks:
Adds 3 properties to analysis that are useful for services:

1. `isDirectlyInvoked` tells us whether a service is either the root script, or there's a path from the root script to the service that only passes through no-command scripts (which counts the same way). This is useful because being directly invoked means you should only exit when wireit exits; instead of when all scripts that depend on you have finished.

2. `services` lists the services that must be started before a script can start. This isn't quite as simple as "my dependencies which are services", again because of needing to traverse through no-command scripts.

3. `serviceConsumers` is the reverse of `services`. It's useful for services to know which scripts could *potentially* depend on a script, because once we start a service, we don't want to shut it down until we know that every depending script has either finished, or didn't need to run at all. Knowing how many to expect from the start makes this simpler.

Also adds a unit test that invokes the `Analyzer` directly. All of our other tests are integration, but in this case it's useful to be able to check the analysis result directly.

Also noticed our uvu test patterns were effectively suffix matches, instead of exact matches (I noticed because `analysis\.test\.js$` was accidentally matching both `analysis.test.js` and `errors-analysis.test.js`).

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -132,7 +132,11 @@ export class Analyzer {
       }
       // We don't care about the result, if there's a cycle error it'll
       // be added to the scripts' diagnostics.
-      this._checkForCyclesAndSortDependencies(info.placeholder, new Set());
+      this._checkForCyclesAndSortDependencies(
+        info.placeholder,
+        new Set(),
+        true
+      );
     }
 
     return this._getDiagnostics();
@@ -196,7 +200,8 @@ export class Analyzer {
     }
     const cycleResult = this._checkForCyclesAndSortDependencies(
       rootConfig,
-      new Set()
+      new Set(),
+      true
     );
     if (!cycleResult.ok) {
       return {
@@ -939,7 +944,8 @@ export class Analyzer {
    */
   private _checkForCyclesAndSortDependencies(
     config: LocallyValidScriptConfig | ScriptConfig | InvalidScriptConfig,
-    trail: Set<ScriptReferenceString>
+    trail: Set<ScriptReferenceString>,
+    isDirectlyInvoked: boolean
   ): Result<ScriptConfig, InvalidScriptConfig> {
     if (config.state === 'valid') {
       // Already validated.
@@ -1059,7 +1065,12 @@ export class Analyzer {
         }
         const result = this._checkForCyclesAndSortDependencies(
           dependency.config,
-          trail
+          trail,
+          // Walk through no-command scripts when determining if something is
+          // being directly invoked (e.g. if the top-level script has no command
+          // and simply delegates to one or more other scripts, then those
+          // dependencies are effectively being directly invoked).
+          isDirectlyInvoked && config.command === undefined
         );
         if (!result.ok) {
           return {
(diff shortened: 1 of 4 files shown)


src/analyzer.ts
```diff
@@ -939,7 +944,8 @@ export class Analyzer {
    */
   private _checkForCyclesAndSortDependencies(
     config: LocallyValidScriptConfig | ScriptConfig | InvalidScriptConfig,
-    trail: Set<ScriptReferenceString>
+    trail: Set<ScriptReferenceString>,
```
justinfagnani, line 948: Nit: Might be a nicer callsite if you gave `trail` and `isDirectlyInvoked` default values, like `trail: Set<ScriptReferenceString> = new Set()`. Especially `trail`, but it also looks like `isDirectlyInvoked` is only potentially false when it's called recursively.

src/analyzer.ts
```diff
@@ -1063,21 +1064,36 @@ export class Analyzer {
...
+              config,
+              validDependencyConfigResult.error.dependencyFailure
+            ),
           };
         }
+        const validDependencyConfig = validDependencyConfigResult.value;
+        if (validDependencyConfig.service) {
+          // We directly depend on a service.
+          config.effectiveServiceDependencies.push(validDependencyConfig);
+        } else if (validDependencyConfig.command === undefined) {
```
justinfagnani, line 1090: So effectiveServiceDependencies do not include transitive service dependencies, except in the no-command case?

src/config.ts
```diff
@@ -85,6 +85,16 @@ export interface ServiceScriptConfig
    * Whether this service is being invoked directly (e.g. `npm run serve`).
    */
   isDirectlyInvoked: boolean;
+
+  /**
+   * Which scripts depend on this service.
+   *
+   * "Effective" meaning these are not necessarily our direct dependents, since
+   * we include transitive service dependencies through no-command scripts.
```
justinfagnani, line 93: Can you comment on which order these are in since you use "reversed"? I presume this is in deepest dependencies first? And do you need to store it reversed? Can you reverse the array at the use site?

aomarks: The order in the array doesn't matter. This is reverse in the sense of the direction of the edges in the build graph; it's the opposite direction of `effectiveServiceDependencies`.

- `effectiveServiceDependencies`: The services I depend on
- `reverseEffectiveServiceDependencies`: The services that depend on me

I thought about using `dependency` vs `dependent` to distinguish, but I find that it's really easy to misread those.

Any other suggestions for the names?

src/analyzer.ts
aomarks: Yeah, exactly.

We traverse through no-command scripts when doing this analysis. This is true pretty generally when thinking about the build graph, because a no-command script is basically just a shorthand alias for a collection of other scripts.

src/analyzer.ts
aomarks: I actually prefer not having a default, because it forces you to think about what the value should be at every call site. It might be easy to just forget to include it at all, otherwise.

src/config.ts
justinfagnani: ohhh! Sorry. Yeah, reverse is a little confusing to me. `dependent` might be better even if it's visually close because it's semantically clearer. I would probably default to that unless something clearly better is found. Maybe something with "back" like a back reference would work?

justinfagnani (approved):

src/config.ts
aomarks: Renamed to `services` and `serviceConsumers`

PR #471, opened 2022-10-20
Services: Refactoring to surface more Execution API to other scripts

aomarks:
A refactoring which allows scripts to directly access full `Execution` instances, instead of just fingerprints as before.

This is useful for the lazy startup/shutdown feature of services, because other scripts need to not only get the service's fingerprint, but also to have a `start` method they can call once they eventually decide whether they need to run or not.

Also:

- Upgraded the self-signed TLS certificate we use in tests, because after upgrading my Linux distro and/or my Node version, these tests were failing locally due to the key being too weak.

- Renamed `script` to `config` in a few places to remove ambiguity.

- Use `_foo` style for `protected` methods.

- Enabled `noImplicitOverride` tsc flag

### DIFF
--- a/src/execution/service.ts
+++ b/src/execution/service.ts
@@ -8,23 +8,27 @@ import {BaseExecution} from './base.js';
 import {Fingerprint} from '../fingerprint.js';
 
 import type {ExecutionResult} from './base.js';
-import type {Executor} from '../executor.js';
 import type {ServiceScriptConfig} from '../config.js';
 import type {Logger} from '../logging/logger.js';
+import type {Executor} from '../executor.js';
 
 /**
  * Execution for a {@link ServiceScriptConfig}.
  */
 export class ServiceScriptExecution extends BaseExecution<ServiceScriptConfig> {
   static execute(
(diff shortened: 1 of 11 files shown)


src/execution/service.ts
```diff
@@ -8,23 +8,27 @@ import {BaseExecution} from './base.js';
...
 import type {Logger} from '../logging/logger.js';
+import type {Executor} from '../executor.js';
 
 /**
  * Execution for a {@link ServiceScriptConfig}.
  */
 export class ServiceScriptExecution extends BaseExecution<ServiceScriptConfig> {
   static execute(
-    config: ServiceScriptConfig,
+    script: ServiceScriptConfig,
```
justinfagnani: did you mean to go back to `script`?

justinfagnani (approved):

src/execution/service.ts
aomarks: Oops, no, merge mistake. But this whole method is removed in the next commit anyway so no point fixing.

PR #473, opened 2022-10-21
Services: Track when services are not needed

aomarks:
Adds a `servicesNotNeeded` promise, which gets resolved when a script either knows it will never run, or when it has finished running, and hence no longer needs the services it depends on to be running. This is how services will keep track of how many dependents still need them to be running -- once there are none left, they can shut down (unless they are directly invoked).

This is done via a new `BaseExecutionWithCommand` base class, which is only relevant to scripts with commands (because no-command scripts don't directly consume services).

Also:

- Pass `abort` promise down to services. Directly-invoked services will need this to know when to shut down on SIGINT.

- Removed a redundant `throw` I noticed.

### DIFF
--- a/src/execution/standard.ts
+++ b/src/execution/standard.ts
@@ -60,57 +60,61 @@ export class StandardScriptExecution extends BaseExecutionWithCommand<StandardSc
   }
 
   protected async _execute(): Promise<ExecutionResult> {
-    this._ensureState('before-running');
+    try {
+      this._ensureState('before-running');
 
-    const dependencyFingerprints = await this._executeDependencies();
-    if (!dependencyFingerprints.ok) {
-      dependencyFingerprints.error.push(this._startCancelledEvent);
-      return dependencyFingerprints;
-    }
-
-    // Significant time could have elapsed since we last checked because our
-    // dependencies had to finish.
-    if (this._shouldNotStart) {
-      return {ok: false, error: [this._startCancelledEvent]};
-    }
-
-    return this._acquireSystemLockIfNeeded(async () => {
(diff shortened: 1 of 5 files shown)


src/execution/standard.ts
```diff
@@ -60,57 +60,61 @@ export class StandardScriptExecution extends BaseExecution<StandardScriptConfig>
...
-      dependencyFingerprints.error.push(this._startCancelledEvent);
-      return dependencyFingerprints;
-    }
-
-    // Significant time could have elapsed since we last checked because our
-    // dependencies had to finish.
-    if (this._shouldNotStart) {
-      return {ok: false, error: [this._startCancelledEvent]};
-    }
+    try {
```
aomarks, line 63: Note this is not a very well formatted diff. It's all indentation inside the `try`.

src/executor.ts
```diff
@@ -140,7 +142,12 @@ export class Executor {
       if (config.command === undefined) {
         execution = new NoCommandScriptExecution(config, this, this._logger);
       } else if (config.service) {
-        execution = new ServiceScriptExecution(config, this, this._logger);
+        execution = new ServiceScriptExecution(
+          config,
+          this,
+          this._logger,
+          this._abort.promise
```
justinfagnani, line 149: Why aren't you using an AbortSignal?

justinfagnani (approved):

src/executor.ts
aomarks: Because it's only supported in Node 15+, and we still support Node 14 until it goes out of LTS next year. Simpler than having a polyfill since it's so trivial anyway.

PR #474, opened 2022-10-21
Services: Integrate into standard scripts

aomarks:
Integrates services into standard scripts (though note services don't yet actually _do_ anything, that's coming up next). Summary:

1. Before a standard script runs, all of its services must have started. If any service failed to start, we fail.

2. While a standard script is still running, if any of its services unexpectedly shuts down, then we're in an invalid state and fail too.

### DIFF
--- a/src/execution/base.ts
+++ b/src/execution/base.ts
@@ -104,6 +104,16 @@ export abstract class BaseExecutionWithCommand<
   readonly servicesNotNeeded = this._servicesNotNeeded.promise;
 
   /**
+   * Resolves when any of the services this script depends on have terminated
+   * (see {@link ServiceScriptExecution.terminated} for exact definiton).
+   */
+  readonly anyServiceTerminated = Promise.race(
+    this._config.services.map(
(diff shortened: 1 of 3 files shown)


src/execution/base.ts
```diff
@@ -102,4 +102,37 @@ export abstract class BaseExecutionWithCommand<
...
+      (service) => this._executor.getExecution(service).terminated
+    )
+  );
+
+  /**
+   * Ensure that all of the services this script depends on are running.
+   */
+  protected async _startServices(): Promise<Result<void, Failure[]>> {
+    if (this._config.services.length > 0) {
+      const results = await Promise.all(
```
justinfagnani, line 121: Is `Promise.all()` ok here (over `Promise.allSettled()`) because you have your own error results that resolved to, not rejected?

justinfagnani (approved):

src/execution/base.ts
aomarks: Yeah we throw only as a panic for an internal error, so in general we don't need to catch exceptions.

PR #475, opened 2022-10-22
Services: State diagram

aomarks:
Adds a state diagram as documentation for services.

### DIFF
--- a/src/execution/service.ts
+++ b/src/execution/service.ts
@@ -17,6 +17,62 @@ import type {Result} from '../error.js';
 
 /**
  * Execution for a {@link ServiceScriptConfig}.
+ *
+ * ```
+ *                    ┌─────────┐
+ *     ╭─◄─ abort ────┤ INITIAL │
+ *     │              └────┬────┘
+ *     │                   │
+ *     ▼                execute
+ *     │                   │
+ *     │           ┌───────▼────────┐
+ *     ├─◄─ abort ─┤ FINGERPRINTING ├──── depExecErr ────►───╮
+ *     │           └───────┬────────┘                        │
+ *     │                   │                                 │
+ *     ▼             fingerprinted                           │
+ *     │                   │                                 │
+ *     │             ┌─────▼─────┐                           │
+ *     ├─◄─ abort ───┤ UNSTARTED │                           │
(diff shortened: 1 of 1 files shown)


src/execution/service.ts
```diff
@@ -17,6 +17,62 @@ import type {Result} from '../error.js';
...
+ *     ▼    │              │  ╭─╮            │           │   │
+ *     │    │              │  │ start        │           │   │
+ *     │    │         ┌────▼──▼─┴┐           │           │   │
+ *     │    ╰─────────► STOPPING ◄─────────────◄─────────╯   │
+ *     │              └┬─▲─┬─────┘           │               │
+ *     │           abort │ │                 │               │
+ *     │               ╰─╯ │                 │               │
+ *     │                  exit               │               │
+ *     │                   │ ╭─╮             │               │ ╭─╮
+ *     │                   │ │ start         │               │ │ start
```
justinfagnani, line 69: Why do the various start events loop back rather than restart the service? Is this a single execution instance so a start event is also handled/generated at another layer that's starting a new execution instance?

aomarks: Yeah, the service class represents a service *bound to a particular execution* of the script graph.

When you `npm run`, there is just one execution. We never need to restart with `npm run`, because we just keep the service running until it's no longer needed.

When you `npm run --watch`, there is one execution *per watch iteration*. The way services will work in watch mode is that at the end of an iteration, any directly-invoked services will become *detached*, meaning the child process will be transferred to a new execution instance.

aomarks: (Updated the comment with more detail, in case you'd read the first version)

aomarks: Also to clarify why `start` would get called multiple times: every script that consumes a service makes its own call to `start`. Only the first one actually causes the process to start up.

justinfagnani: Thanks for the explanation! Is that described anywhere else? Might good to link from here if not.

justinfagnani (approved):

src/execution/service.ts
aomarks: Added a comment to this class

PR #476, opened 2022-10-24
Services: Support for very basic consumer/service test case

aomarks:
Just a start on the service state machine, with a test. Handling for the most basic simple case: one standard consumer that depends on one service. Obviously many cases not yet handled/tested here, but will build out more complexity incrementally.

Also fixes a bug with a missing `await`, and adds a small test utility method.

### DIFF
--- a/src/execution/service.ts
+++ b/src/execution/service.ts
@@ -10,7 +10,7 @@ import {Deferred} from '../util/deferred.js';
 import {ScriptChildProcess} from '../script-child-process.js';
 
 import type {ExecutionResult} from './base.js';
-import type {ServiceScriptConfig} from '../config.js';
+import type {ScriptReference, ServiceScriptConfig} from '../config.js';
 import type {Executor} from '../executor.js';
 import type {Logger} from '../logging/logger.js';
 import type {Failure} from '../event.js';
@@ -18,7 +18,14 @@ import type {Result} from '../error.js';
 
 type ServiceState =
   | {id: 'initial'}
-  | {id: 'fingerprinting'}
+  | {
+      id: 'executingDeps';
+      fingerprint: Deferred<ExecutionResult>;
+    }
+  | {
+      id: 'fingerprinting';
+      fingerprint: Deferred<ExecutionResult>;
+    }
   | {id: 'unstarted'}
   | {
       id: 'starting';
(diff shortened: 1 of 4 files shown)


src/execution/standard.ts
```diff
@@ -75,7 +75,7 @@ export class StandardScriptExecution extends BaseExecutionWithCommand<StandardSc
         return {ok: false, error: [this._startCancelledEvent]};
       }
 
-      return this._acquireSystemLockIfNeeded(async () => {
+      return await this._acquireSystemLockIfNeeded(async () => {
```
justinfagnani, line 78: This is a bit subtle!

src/test/util/test-rig-command.ts
```diff
@@ -194,6 +194,13 @@ export class WireitTestRigCommandInvocation extends IpcClient<
     return this._closed.promise;
   }
 
+  /**
+   * Return whether this invocation is still running.
+   */
+  get running(): boolean {
```
justinfagnani, line 200: What's our style guide say for boolean fields? Should this be `isRunning`?

src/execution/service.ts
```diff
@@ -15,6 +16,32 @@ import type {Logger} from '../logging/logger.js';
...
+  | {
+      id: 'starting';
+      child: ScriptChildProcess;
+      started: Deferred<Result<void, Failure[]>>;
+    }
+  | {
+      id: 'started';
+      child: ScriptChildProcess;
+    }
+  | {id: 'stopping'}
```
justinfagnani, line 32: Does this state still have an associated child process?

src/execution/service.ts
```diff
@@ -109,22 +137,148 @@ export class ServiceScriptExecution extends BaseExecutionWithCommand<ServiceScri
...
+  private _onChildStarted() {
+    switch (this._state.id) {
+      case 'starting': {
+        this._state.started.resolve({ok: true, value: undefined});
+        this._logger.log({
+          script: this._config,
+          type: 'info',
+          detail: 'service-started',
+        });
+        this._state = {
```
justinfagnani, line 210: Should the state transition after the promises below settle?

src/execution/service.ts
```diff
@@ -109,22 +137,148 @@ export class ServiceScriptExecution extends BaseExecutionWithCommand<ServiceScri
...
+        this._state = {id: 'fingerprinting'};
+        const dependencyFingerprints = await this._executeDependencies();
+        if (!dependencyFingerprints.ok) {
+          return dependencyFingerprints;
+        }
+        const fingerprint = await Fingerprint.compute(
+          this._config,
+          dependencyFingerprints.value
+        );
+        this._state = {id: 'unstarted'};
```
justinfagnani, line 151: I think it's a little confusing to have the state set twice in this case. It's also a bit confusing to see some of these functions be async and use await and others use unawaited promises to set up the next transition... along with the boilerplate of the non-relevant states, I'd maybe in the future consider a state machine library, or some helpers to make the machine even more declarative.

justinfagnani (approved):
👍

src/execution/service.ts
aomarks: True. Replaced the `async` function with the event method style. This does require another state, but it makes sense. (The one other `async` function was just a mistake, it didn't actually `await`).

On the boilerplate: as this class grows, a lot more cases end up being meaningful. For now, I like that it forces me to consider every possible state. In some cases there will end up being only one valid state, so that could just be replaced with a simple assert at the top. Keeping it all consistent for now though.

PR #488, opened 2022-10-28
Services: Pass running services across watch iterations

aomarks:
Implements the basic logic for passing running services across watch mode iterations:

1. The top-level `execute` method now returns a map of the services that are still running at the end of the execution.

2. We pass that service map to the next iteration, and pass the previous version of each service into its new version.

3. As soon as a service knows its fingerprint, we check if it matches the previous version's fingerprint. If it does match, we "adopt" it instead of starting a new process. If it does not match, we shut it down, and then continue as normal (effectively restarting it).

Still a few cases to handle here, which are in TODOs, but will do in followup PRs.

### DIFF
--- a/src/execution/service.ts
+++ b/src/execution/service.ts
@@ -20,23 +20,33 @@ type ServiceState =
   | {
       id: 'initial';
       entireExecutionAborted: Promise<void>;
+      adoptee: ServiceScriptExecution | undefined;
     }
   | {
       id: 'executingDeps';
-      fingerprint: Deferred<ExecutionResult>;
+      deferredFingerprint: Deferred<ExecutionResult>;
+      adoptee: ServiceScriptExecution | undefined;
     }
   | {
       id: 'fingerprinting';
-      fingerprint: Deferred<ExecutionResult>;
+      deferredFingerprint: Deferred<ExecutionResult>;
+      adoptee: ServiceScriptExecution | undefined;
+    }
+  | {
+      id: 'stoppingAdoptee';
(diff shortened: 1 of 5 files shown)


aomarks:
Rebased (sorry if was already reviewing)

src/executor.ts
```diff
@@ -118,7 +124,11 @@ export class Executor {
   /**
    * Execute the root script.
    */
-  async execute(): Promise<Result<void, Failure[]>> {
+  async execute(): Promise<Result<ServiceMap, Failure[]>> {
+    // TOOD(aomarks) If we have any running services from a previous watch
```
justinfagnani, line 128: Do you plan on keeping services running across edits tot he wireit configs?

justinfagnani (approved):

src/executor.ts
aomarks: Yep, as long as the fingerprint stays the same for a service, we can keep it running. So if you change something that would require the service to restart, it will do so. Otherwise no need to restart it.

src/execution/service.ts
```diff
@@ -194,6 +200,34 @@ export class ServiceScriptExecution extends BaseExecutionWithCommand<ServiceScri
     };
   }
 
+  /**
+   * Return the fingerprint of this service. If the fingerprint is not yet
+   * computed, or if the service is stopped/failed/detached, returns undefined.
```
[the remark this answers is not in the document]
aomarks: Yeah, we probably should. Done.

PR #489, opened 2022-10-28
Services: Shut down services that were deleted between watch iterations

aomarks:
Handles the case where we're in watch mode, and a service was entirely deleted from the graph. Previously it would keep running, but now we notice it and shut it down before starting the next execution. The same goes for scripts that used to be "directly invoked" and are now no longer (since those might not need to run at all).

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -125,10 +125,32 @@ export class Executor {
    * Execute the root script.
    */
   async execute(): Promise<Result<ServiceMap, Failure[]>> {
-    // TOOD(aomarks) If we have any running services from a previous watch
-    // iteration, we should at this point shut down any of the ones that have
-    // since been deleted from the build graph entirely, or which have become
-    // non-directly-invoked.
+    if (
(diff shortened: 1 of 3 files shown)


src/executor.ts
```diff
@@ -125,10 +125,32 @@ export class Executor {
...
+          currentDirectlyInvokedServices.add(scriptReferenceToString(script));
+        }
+      }
+      const stopPromises = [];
+      for (const [key, service] of this._previousIterationServices) {
+        if (!currentDirectlyInvokedServices.has(key)) {
+          const child = service.detach();
+          if (child !== undefined) {
+            child.kill();
+            stopPromises.push(child.completed);
```
justinfagnani, line 146: can `child.completed` reject? If so use `Promise.allSettled()` below.

justinfagnani (approved):

src/executor.ts
aomarks: It cannot

PR #496, opened 2022-10-30
Trying out codeql action

aomarks:
Code scanning for common bugs provided by GitHub: https://codeql.github.com/docs/codeql-overview/about-codeql/

See https://github.com/github/codeql/tree/main/javascript/ql/src for the list of checks.


justinfagnani:
I'm kind of confused by the description as to what this actually does. Is there a default security analysis in addition to the code querying?

aomarks:
> I'm kind of confused by the description as to what this actually does. Is there a default security analysis in addition to the code querying?

Yep, you can find the default set of checks at https://github.com/github/codeql/tree/main/javascript/ql/src

I added a couple errors to this CL temporarily that should get caught. See above for example. One of them was for jsdoc (which didn't fire for some reason) and one of them for weak password hashing, which did!

aomarks:
> I added a couple errors to this PR temporarily that should get caught. See above for example. One of them was for bad jsdoc (which didn't fire for some reason) and one of them for weak password hashing, which did!

For the record, in case removing the error makes this go away:

image

justinfagnani:
Awesome, thanks for triggering the demo @aomarks ! I wish their docs page lead to a demo. Maybe I missed it.

justinfagnani (approved):

aomarks:
> Awesome, thanks for triggering the demo @aomarks ! I wish their docs page lead to a demo. Maybe I missed it.

I had to dig for it, the docs site is not super clear I agree.

PR #497, opened 2022-10-31
Services: Fix issue with caching in watch mode due to missing promise resolution

aomarks:
I've started testing services for real on webcomponents.org, and I hit a bug where watch mode could get stuck because of some missing promise resolution.

### DIFF
--- a/src/execution/service.ts
+++ b/src/execution/service.ts
@@ -746,6 +746,8 @@ export class ServiceScriptExecution extends BaseExecutionWithCommand<ServiceScri
       case 'unstarted':
       case 'depsStarting': {
         this._state = {id: 'stopped'};
+        this._terminated.resolve({ok: true, value: undefined});
+        this._servicesNotNeeded.resolve();
         return;
       }
       case 'stopping':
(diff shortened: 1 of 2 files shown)


src/execution/service.ts
```diff
@@ -746,6 +746,8 @@ export class ServiceScriptExecution extends BaseExecutionWithCommand<ServiceScri
       case 'unstarted':
       case 'depsStarting': {
         this._state = {id: 'stopped'};
+        this._terminated.resolve({ok: true, value: undefined});
```
justinfagnani, line 749: Why don't these need to be resolved above?

If these are pre-conditions for entering the stopped state, it'd be nice if they could be checked / enforced there. That would require abstracting over entering states so that you don't just do:

```ts
this._state = {id: 'stopped'};
```

but

```ts
this._enterStoppedState();
```

etc

justinfagnani (approved):

src/execution/service.ts
aomarks: The above cases are going to the `stopped` state. We don't enter `stopped` until the child is killed (or never starts).

I do like the `_enterStoppedState` idea though, since there is another spot above where we enter that state. Also renamed `_fail` to `_enterFailedState` to match.

aomarks: (Edit: typo)

PR #504, opened 2022-11-03
Services: Fix memory leaks in watch mode

aomarks:
Fixes two memory leaks that affected watch mode:

1. An `abort` promise which always remained unresolved until wireit exited. This prevented both standard and service executions from ever being garbage collected, because they both awaited those promises indefinitely, even after those instances were stale.

    The fix was to replace the promise entirely with explicit `abort()` methods, which turned out to be simpler and less error-prone anyway.

2. The `Executor` for watch mode iteration N was holding references to all persistent services from watch mode iteration N-1, and so on all the way back through all iterations, because of the `previousIterationServices` map that we pass forward across iterations.

    The fix was to delete services from the map after we know they are adopted, breaking the reference chain.

This PR includes new garbage collection tests that use `FinalizationRegistry` to keep track of how many instances of `Executors` and `Executions` there are after a `global.gc()`. I'm just running that on a single OS/node version on CI, since I think it might be slightly flaky, plus I think that's good enough anyway.

In writing the GC tests, I also found a few other bugs with services -- some unhandled state transitions, and a timing issue where we weren't waiting for persistent services to be ready before indicating that the first phase of execution was done.

### DIFF
--- a/src/test/gc.test.ts
+++ b/src/test/gc.test.ts
@@ -8,7 +8,11 @@ import {suite} from 'uvu';
 import * as assert from 'uvu/assert';
 import {timeout} from './util/uvu-timeout.js';
 import {WireitTestRig} from './util/test-rig.js';
-import {Executor, registerExecutorConstructorHook} from '../executor.js';
+import {
+  Executor,
+  registerExecutorConstructorHook,
+  ServiceMap,
+} from '../executor.js';
 import {Analyzer} from '../analyzer.js';
 import {DefaultLogger} from '../logging/default-logger.js';
 import {WorkerPool} from '../util/worker-pool.js';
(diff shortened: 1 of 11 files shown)


src/test/gc.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    global.gc();
+    try {
+      cb();
+      return;
+    } catch {
+      // Ignore
+    }
+    await new Promise((resolve) => setTimeout(resolve, wait));
+  }
+  cb();
```
justinfagnani, line 79: why this extra call?

src/test/gc.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      const result = await resultPromise;
+      if (!result.ok) {
+        for (const error of result.error) {
+          logger.log(error);
+        }
+        throw new Error(`Execution error`);
+      }
+    }
+
+    await retryWithGcUntilCallbackDoesNotThrow(() => {
```
justinfagnani, line 136: how often does this need all the retries to pass?

justinfagnani (approved):

src/test/gc.test.ts
aomarks: In a single run I just did, one time needed `10`, another needed `100`.

src/test/gc.test.ts
aomarks: Added a comment. It's the attempt after the final wait, where we don't wrap in a `try` so that the exception bubbles up.

Issue #330, opened 2022-06-28
GitHub caching should recover from TCP timeouts


aomarks:
Sometimes we get errors like this from GitHub caching:

```
❌ [test:cache-github] Internal error! Please file a bug at https://github.com/google/wireit/issues/new, mention this message, that you encountered it in wireit version 0.7.1, and give information about your package.json files.
    Unknown error thrown: Error: connect ETIMEDOUT 13.107.42.16:443
Error: connect ETIMEDOUT 13.107.42.16:443
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1159:16)
```

We should recover from these the same as we do from 429s and 503s.

aomarks:
Similar error we should also handle here:

```
❌ [packages/labs/gen-utils:test] Internal error! Please file a bug at https://github.com/google/wireit/issues/new, mention this message, that you encountered it in wireit version 0.7.1, and give information about your package.json files.
    Unknown error thrown: Error: read ECONNRESET
Error: read ECONNRESET
    at TLSWrap.onStreamRead (node:internal/stream_base_commons:217:20)
```

rictic:
Still seeing this regularly

PR #510, opened 2022-11-07
Recover from ECONNRESET and similar GitHub caching connection errors

aomarks:
This should handle ECONNRESET and hopefully also ETIMEDOUT, though the latter is a little trickier to reproduce in a unit test.

Possibly fixes https://github.com/google/wireit/issues/330

### DIFF
--- a/src/caching/github-actions-cache.ts
+++ b/src/caching/github-actions-cache.ts
@@ -120,26 +120,26 @@ export class GitHubActionsCache implements Cache {
 
     const {req, resPromise} = this._request(url);
     req.end();
-    const res = await resPromise;
+    const result = await resPromise;
+    if (this._maybeHandleServiceDown(result, script)) {
+      return undefined;
+    }
+    const response = result.value;
 
-    if (res.statusCode === /* No Content */ 204) {
+    if (response.statusCode === /* No Content */ 204) {
       return undefined;
     }
 
-    if (isOk(res)) {
-      const {archiveLocation} = JSON.parse(await readBody(res)) as {
+    if (isOk(response)) {
+      const {archiveLocation} = JSON.parse(await readBody(response)) as {
         archiveLocation: string;
       };
       return new GitHubActionsCacheHit(script, archiveLocation);
     }
 
-    if (this._maybeHandleServiceDown(res, script)) {
-      return undefined;
-    }
-
     throw new Error(
-      `GitHub Cache check HTTP ${String(res.statusCode)} error: ` +
-        (await readBody(res))
+      `GitHub Cache check HTTP ${String(response.statusCode)} error: ` +
+        (await readBody(response))
     );
   }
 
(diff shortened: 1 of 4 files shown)


src/caching/github-actions-cache.ts
```diff
@@ -345,33 +351,48 @@ export class GitHubActionsCache implements Cache {
    * Otherwise return false.
    */
   private _maybeHandleServiceDown(
-    res: http.IncomingMessage,
+    res: Result<http.IncomingMessage, Error>,
     script: ScriptReference
-  ): boolean {
+  ): res is {ok: false; error: Error} {
     if (this._serviceIsDown) {
       return true;
```
rictic, line 358: in this case do we know that `res` is `{ok: false, error: Error}`?

rictic (approved):

src/caching/github-actions-cache.ts
```diff
@@ -345,33 +351,48 @@ export class GitHubActionsCache implements Cache {
...
+          `Connection error from GitHub Actions service, caching disabled. ` +
+          'Detail: ' +
+          ('code' in res.error
+            ? `${(res.error as Error & {code: string}).code} `
+            : '') +
+          res.error.message,
+      });
+    } else {
+      switch (res.value.statusCode) {
+        default: {
```
justinfagnani, line 375: I find `default` first pretty unusual. Our JS style guide says it should come last too, fwiw.

src/caching/github-actions-cache.ts
```diff
@@ -345,33 +351,48 @@ export class GitHubActionsCache implements Cache {
...
+          return false;
+        }
+        case /* Too Many Requests */ 429: {
+          this._logger.log({
+            script,
+            type: 'info',
+            detail: 'generic',
+            message: `Hit GitHub Actions cache rate limit, caching disabled.`,
+          });
+          break;
```
justinfagnani, line 385: Similar comment to @rictic, but these cases seem to contradict the function's assertion type. Here we know that `res.ok` is truthy, but we're asserting that it's `false`. Can the assertion omit `ok` or otherwise be tightened up?

justinfagnani (changes requested):

src/caching/github-actions-cache.ts
aomarks: Moved to end

src/caching/github-actions-cache.ts
aomarks: No, good point. It was also slightly wrong for the other cases, since we also handled the case where ok was true, but there were certain HTTP error codes.

Removed this case, since it was only there to prevent logging, and inverted the return value so that we can correctly say that if we return `true`, then `ok` must be true (which is true!).

src/caching/github-actions-cache.ts
aomarks: Fixed, see reply to @rictic.

aomarks: Fixed, see reply to @rictic.

justinfagnani (approved):

PR #519, opened 2022-11-08
Use latest pre-release version for our own tests

aomarks:
In particular I want to start seeing whether https://github.com/google/wireit/pull/510 is having an effect


augustjk (approved):

aomarks:
Wow no failures, first time that's happened in a long time.

PR #520, opened 2022-11-08
Rename triggersRerun to cascade, and add new documentation

aomarks:
First it was `soft`, then it was `triggersRerun`, now it's `cascade` (note we haven't released yet, so none of these changes were breaking).

This name came to me last night, and I kind of like it. I think `soft` is too vague and non-technical, but `triggersRerun` is just a mouthful (I've been typing it out a bunch, and it's been getting on my nerves). `cascade` feels like it sort of gives an impression of what the thing does, and I think will be memorable and understandable once you've read the docs.

Speaking of, I also rewrote the docs in a way that I hope is more clear.

Also adds some missing docs about `dependencies` and `service` to the configuration reference table.

Addon to https://github.com/google/wireit/issues/237


README.md
```diff
@@ -521,6 +447,118 @@ If you have a service that produces output, you should define a _non-service_
...
+## Execution cascade
+
+By default, a script always needs to run (or restart in the case of
+[`services`](#services)) if any of its dependencies needed to run, _regardless
+of whether the dependency produced new or relevant output_.
+
+This automatic _cascade_ of script execution is the default behavior because it
+ensures that any _possible_ output produced by a script propagates to all other
+scripts that might depend on it. In other words, Wireit does not assume that the
+`files` array completely describes the inputs to a script.
```
augustjk, line 459: These sentences are a bit hard to understand because the context around which script are we talking about. The script that runs as a dependency, or the script who has the dependency.
```suggestion
This automatic _cascade_ of script execution is the default behavior because it
ensures that any _possible_ output produced by a dependent script propagates to all other
scripts that might depend on it. In other words, Wireit does not assume that the
`files` array completely describes the inputs to a script with dependencies.
```

augustjk (approved):
Also do we need to make sure the docs site is synced with the readme?

aomarks:
> Also do we need to make sure the docs site is synced with the readme?

Yeah I haven't been keeping it up to date. Never publicized it, was kinda a work in progress. I've taken it down for now.

PR #523, opened 2022-11-10
Code line prettified

ra1nbow1:


src/error.ts
```diff
@@ -55,7 +55,7 @@ export class DiagnosticPrinter {
 
     const path = this._formatPath(diagnostic.location);
     let result = `❌ ${path} ${diagnostic.message}
-${drawSquiggle(diagnostic.location, 4)}`;
+    ${drawSquiggle(diagnostic.location, 4)}`;
```
aomarks, line 58: I think the previous indentation was intentional, because we don't want to print that indentation on the terminal when this diagnostic is displayed.

aomarks:
Thank you for the PR! However mentioned in the comment, I think we want to preserve the existing indentation in this case.

ra1nbow1:
> Thank you for the PR! However mentioned in the comment, I think we want to preserve the existing indentation in this case.

Okay, no problem :)

PR #526, opened 2022-11-12
Services: Don't fail in watch mode when a dependency restarts or fails

aomarks:
In non-watch mode, if service A depends on service B, and service B exits unexpectedly, then service A is automatically killed. We do this because otherwise wireit will just continue running, even though there has been an error. We instead want to exit with a non-zero code.

Before this PR, we did the same thing in watch mode. However, this was wrong for 2 reasons:

1. If we're iteration N-1 which is about to be adopted into iteration N, our dependencies will sometimes *intentionally* restart, but we were treating this as though it was unexpected. This should not cause us to fail, since we'll either also restart very shortly (when cascade is true), or we'll just keep running (when cascade is false).
          
2. If we're iteration N and our dependency unexpectedly exits by itself, it's not actually helpful if we also aggressively exit. We can just keep running, and hopefully the user can fix the problem with the server dependency, which will then either cause us to restart or not depending on the cascade setting.

As part of this I found these other related issues, which were compounding the problem above and making it harder to debug:

- We were returning `undefined` when we detached, if the service was in its "failing" or "stopping" state, instead of the child process. This was compounding the issue described above, because if iteration N-1 was failing, iteration N would think the child wasn't running *at all*, and would not wait for it to fully shut down. We are now much stricter about communicating from iteration N-1 to iteration N that there is a running child process, regardless of the state.

- We didn't log when a service shut down because of a failure in its dependencies.

### DIFF
--- a/src/test/service.test.ts
+++ b/src/test/service.test.ts
@@ -1143,4 +1143,82 @@ test(
   })
 );
 
+test(
+  'service in watch mode persists when non-cascading dependency restarts or fails',
+  // service1
+  //    |
+  //    v
+  // service2 (restarts and fails)
+  timeout(async ({rig}) => {
+    const service1 = await rig.newCommand();
+    const service2 = await rig.newCommand();
+    await rig.writeAtomic({
+      'package.json': {
+        scripts: {
+          service1: 'wireit',
+          service2: 'wireit',
+        },
+        wireit: {
+          service1: {
+            command: service1.command,
+            service: true,
+            dependencies: [
+              {
+                script: 'service2',
+                cascade: false,
+              },
+            ],
+            files: ['input/service1'],
+          },
+          service2: {
+            command: service2.command,
+            service: true,
+            files: ['input/service2'],
+          },
+        },
+      },
+    });
+
+    await rig.write('input/service1', '1');
+    await rig.write('input/service2', '1');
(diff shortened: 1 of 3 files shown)


augustjk (approved):
Nice!

src/test/service.test.ts
```diff
@@ -1143,4 +1143,82 @@ test(
...
 
+test(
+  'service in watch mode persists when non-cascading dependency restarts or fails',
+  // service1
+  //    |
+  //    v
+  // service2 (restarts and fails)
+  timeout(async ({rig}) => {
+    const service1 = await rig.newCommand();
+    const service2 = await rig.newCommand();
```
AndrewJakubowicz, line 1154: Unbelievably minor nit and can be discarded. Naming these `parentService` and `childService` could help make it easier to follow the test. I kept losing track of whether `1` or `2` was the child.
Also please ignore this comment.

AndrewJakubowicz (approved):
Great!

src/test/service.test.ts
augustjk: I can +1 that I did have to look at the config being written to tell which one was the dependent. But also, super minor since it is indeed right there.

aomarks: Done

AndrewJakubowicz (approved):

PR #527, opened 2022-11-13
Services: Add `readyWhen` setting to defer until some output is logged

aomarks:
You can now configure a service to only be considered ready after a line on `stdout` or `stderr` has matched a given regular expression:

```json
{
  "command": "node my-server.js",
  "service": {
    "readyWhen": {
      "lineMatches": "Server listening on port \\d+"
    }
  }
}
```


src/util/line-monitor.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  private _onStdout = (data: string | Buffer) => {
+    this._stdout = this._check(this._stdout + String(data));
+  };
+
+  private _onStderr = (data: string | Buffer) => {
+    this._stderr = this._check(this._stderr + String(data));
+  };
+
+  private _check(buffer: string): string {
```
AndrewJakubowicz, line 54: Nit: Could this manual buffering be replaced by using Node's `readline` module, which is supported for Node 14 and up.
E.g.: https://nodejs.org/docs/latest-v14.x/api/readline.html#readline_readline

This then has an event caused `line`: https://nodejs.org/docs/latest-v14.x/api/readline.html#readline_event_line which will provide the single line.

src/fingerprint.ts
```diff
@@ -78,6 +78,14 @@ export interface FingerprintData {
 
   // Must be sorted.
   dependencies: {[dependency: ScriptReferenceString]: FingerprintData};
+
+  service:
```
AndrewJakubowicz, line 82: No action required (or a comment could help). Why doesn't the boolean decision of whether a script is a service not need to be fingerprinted?

src/execution/service.ts
```diff
@@ -776,6 +803,22 @@ export class ServiceScriptExecution extends BaseExecutionWithCommand<ServiceScri
   private _onChildStarted() {
     switch (this._state.id) {
       case 'starting': {
+        if (this._state.readyMonitor !== undefined) {
+          this._state = {
+            id: 'readying',
+            child: this._state.child,
+            fingerprint: this._state.fingerprint,
+            started: this._state.started,
+            readyMonitor: this._state.readyMonitor,
```
AndrewJakubowicz, line 812: What you have written is more precise. But 🏌️:

`this.state = { ...this._state, id: 'readying' };`

src/execution/service.ts
```diff
@@ -876,6 +966,17 @@ export class ServiceScriptExecution extends BaseExecutionWithCommand<ServiceScri
         break;
       }
       case 'starting': {
+        this._state.readyMonitor?.abort();
```
AndrewJakubowicz, line 969: Nit: Does `this._state.child.kill` not need to be called here?

src/test/service.test.ts
```diff
@@ -1221,4 +1221,57 @@ test(
...
+
+    // Logged part of the expected line, but not all of it. Not ready.
+    serviceInv.stdout('Foo\nBar\nListening ');
+    await new Promise((resolve) => setTimeout(resolve, 100));
+    assert.equal(standard.numInvocations, 0);
+
+    // Logged the rest of the expected line. Ready!
+    serviceInv.stdout('on port 8080.\nBaz\n');
+    const standardInv = await standard.nextInvocation();
+
```
AndrewJakubowicz, line 1271: I expected a `numInvocations` assert here. But maybe I didn't understand test fully.

AndrewJakubowicz (approved):
Amazing!

src/execution/service.ts
aomarks: Nope, the way it works that if we haven't started yet, then we go to the `stopping` state, and then in `_onChildStarted` we notice that we are stopped and kill the child. (We can't kill the child before it has started)

src/execution/service.ts
aomarks: I guess what I don't like about that as a general pattern for state transitions is that it will also copy properties from the old state that are not declared in the new state, without a type error.

The worst case would be something like:
1. State A has required property `foo`
2. State B has no property `foo`, but we copy it with the `...`
3. State C has *optional* property `foo`, which *should* be undefined, but we inadvertently propagate the `foo` from A to C using the `...`

(That situation isn't a problem in practice currently, but that's why I prefer just being explicit)

src/fingerprint.ts
aomarks: Because services can never be cached and don't write out a fingerprint file, so there's no bad interaction really possible.

However, we do *now* store whether something is a service, since `service` will only be `undefined` if it is not a service at all.

src/test/service.test.ts
aomarks: We're just checking that we don't start the service until the log has happened. We check that we haven't started with `numInvocations === 0`, and check that we have started with `nextInvocation`. But doesn't hurt to also double check that we only launched once at the end. Added.

src/util/line-monitor.ts
aomarks: I don't think so, because readline only gives you a line when it completed (so only when `\n` is printed do you get the previous line). But we want to check even an incomplete line, whenever any output is emitted.

Because imagine a script that just writes "Listening on port 8080" without writing a trailing newline.

Updated the test to cover that case, and added a comment.

AndrewJakubowicz: Brilliant!

src/execution/service.ts
AndrewJakubowicz: Agree!

src/util/line-monitor.ts
```diff
@@ -12,6 +12,9 @@ import type {Result} from '../error.js';
 /**
  * Monitors the stdout and stderr of a child process line-by-line searching for
  * a match of the given regular expression.
+ *
+ * Note we can't use readline here because we want to check lines that haven't
+ * completed yet.
```
AndrewJakubowicz, line 17: Subtle and brilliant!

AndrewJakubowicz (approved):

augustjk (approved):
🎉

PR #535, opened 2022-11-14
Don't check node/npm version in CI because it's already shown

aomarks:
It's easy to find the node and npm versions from the "Environment details" dropdown under "Run actions/setup-node@v3", so no need for us to check them again. I think I just didn't notice this before.

image


augustjk (approved):
Huh, feels like that's new. Good find!

augustjk:
Ah, it was added last month: https://github.com/actions/setup-node/releases/tag/v3.5.1

PR #537, opened 2022-11-15
Center README logo and tagline

e111077:
I'm an expert CSS person that can center a div

[preview](https://github.com/e111077/wireit/blob/main/README.md)


aomarks (approved):
Looks nice, thanks!

PR #538, opened 2022-11-15
Update extension

rictic:


rictic:
This has been published as v0.4.0

I think in a follow up change we'll want to get a smaller version of the logo, as this one is fairly large, it's the majority of the extension by bytes!

aomarks (approved):
Woohoo

rictic:
Also note for the future: updating the logo locally doesn't do anything?! The logo seems to always be fetched from the marketplace, surprisingly! Even when you've installed the extension from a vsix file!

PR #539, opened 2022-11-15
Hacking-docs

rictic:
This has info on developing and publishing the extension. We don't want this in the main README.md as it would clutter up the marketplace docs.


vscode-extension/HACKING.md
```diff
@@ -0,0 +1,15 @@
+## Developing
+
+Run `npm run install-extension` to build and install the extension in your local VSCode. If you run `npm run install-extension --watch` it will rebuild and reinstall on every change, though you'll need to run `Developer: Reload Window` in VSCode to pick up on the changes.
```
aomarks, line 3: I believe this script is only defined in the vscode-extension subdirectory, so probably should mention that

aomarks (approved):
Nice!

Issue #543, opened 2022-11-17
RangeError: Invalid string length


jorgecasar:
I'm having this bug in version 0.7.2:

```
[path/to/package-30:build:ts] Internal error! Please file a bug at https://github.com/google/wireit/issues/new, mention this message, that you encountered it in wireit version 0.7.2, and give information about your package.json files.
    Unknown error thrown: RangeError: Invalid string length
RangeError: Invalid string length
```

And with version 0.7.3 I'm having this other message:

```
file:///[...]/node_modules/wireit/lib/fingerprint.js:115
            this._str = JSON.stringify(this._data);
                             ^

RangeError: Invalid string length
    at JSON.stringify (<anonymous>)
    at get string [as string] (file:///[...]/node_modules/wireit/lib/fingerprint.js:115:30)
    at LocalCache._getCacheDir (file:///[...]/node_modules/wireit/lib/caching/local-cache.js:49:104)
    at LocalCache.get (file:///[...]/node_modules/wireit/lib/caching/local-cache.js:18:31)
    at file:///[...]/node_modules/wireit/lib/execution/standard.js:66:42
    at async StandardScriptExecution._acquireSystemLockIfNeeded (file:///[...]/node_modules/wireit/lib/execution/standard.js:142:28)
    at async StandardScriptExecution._execute (file:///[...]/node_modules/wireit/lib/execution/standard.js:45:20)

Node.js v18.9.1
```

aomarks:
Thanks for the report! Can you provide the relevant config?

jorgecasar:
I really don't know what is the relevant config because in the previous commit it works fine. I only change typescript files of this package and I included some dependencies in other package. I hope you can understand it, because I can't share the real names.

Here is my config with normalized names. From one commit to the other I only added  `../package-[7-9]:build` in the wireit config of `package-10`. The package-30 is untouched and is the one that generates the error.

package-30:

```
"wireit": {
    "build": {
      "dependencies": [
        "build:ts",
        "build:bundle"
      ]
    },
    "build:bundle": {
      "command": "esbuild",
      "dependencies": [
        "build:ts"
      ],
      "files": [
        "development/**/*.{js,js.map,d.ts,d.ts.map}",
        "esbuild.config.js"
      ],
      "output": [
        "index.js",
        "index.js.map"
      ],
      "packageLocks": [
        "yarn.lock"
      ]
    },
    "build:ts": {
      "command": "tsc --build --pretty",
      "clean": "if-file-deleted",
[...]

PR #546, opened 2022-11-18
Prevent fingerprint size explosion by using hashes of dependency fingerprints

aomarks:
After each script runs, we write its JSON-stringified *fingerprint* to a file, so that the next time it runs we can compare the fingerprint to decide whether it needs to run again or not.

Part of a fingerprint is the fingerprint of its dependencies (this is how what we now describe as the "cascade" behavior works). (The full list of things in the fingerprint is at https://github.com/google/wireit#fingerprint).

Previously, when nesting dependency fingerprints, we nested the *entire data object for that dependency*. The problem with this is that it can cause an explosion of data in the fingerprint, which at a certain point will be too large for V8's `JSON.stringify` function to handle, causing either an `Invalid string length` error or a `heap out of memory` error. This also would have hurt performance, since we were serializing and writing unnecessarily large amounts of data.

This PR fixes that problem. Now, when embedding the fingerprints of dependencies, we instead embed a *hash* of the fingerprint, instead of the whole object. This bounds the size of any given fingerprint dependency to 64 bytes, whereas previously it could grow to any size -- in some configurations expontentially.

Fixes https://github.com/google/wireit/issues/543

### DIFF
--- a/src/fingerprint.ts
+++ b/src/fingerprint.ts
@@ -241,7 +241,6 @@ export class Fingerprint {
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       this._str = JSON.stringify(this._data!) as FingerprintString;
     }
-    console.log(JSON.stringify(this._data, null, 2));
     return this._str;
   }
 
--- a/src/fingerprint.ts
+++ b/src/fingerprint.ts
@@ -61,7 +61,7 @@ export interface FingerprintData {
   clean: boolean | 'if-file-deleted';
 
   // Must be sorted.
-  files: {[packageDirRelativeFilename: string]: Sha256HexDigest};
+  files: {[packageDirRelativeFilename: string]: FileSha256HexDigest};
 
   /**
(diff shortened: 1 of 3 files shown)


src/fingerprint.ts
```diff
@@ -225,12 +234,14 @@ export class Fingerprint {
...
   private _str?: FingerprintString;
   private _data?: FingerprintData;
+  private _hash?: FingerprintSha256HexDigest;
 
   get string(): FingerprintString {
     if (this._str === undefined) {
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       this._str = JSON.stringify(this._data!) as FingerprintString;
     }
+    console.log(JSON.stringify(this._data, null, 2));
```
augustjk, line 244: leftover logging

augustjk (approved):
Looking good except for a lingering log!

src/fingerprint.ts
aomarks: Oops! Fixed, thank you.

PR #548, opened 2022-11-18
Add default excluded folders

aomarks:
The following folders are now excluded by default from both the `files` and `output` arrays:

- `.git/`
- `.hg/`
- `.svn/`
- `.wireit/`
- `CVS/`
- `node_modules/`

In the highly unusual case that a user needs to reference a file in one of those folders, they can set `allowUsuallyExcludedPaths: true` to remove all of these default excludes.

Tracking something inside `node_modules/` is a *somewhat* plausible thing that a user might occasionally want to do.

The reason this didn't come up until recently is that I've always been testing on projects which have e.g. `src/`, `lib/`, `dist/` etc. top-level folders that cleanly partition relevant inputs/outputs. However, I am now trying to integrate with a project that has its source very flat, so the file patterns look like e.g. simply `**/*.js`, and I don't want to have to explicitly write exclude patterns for these folders, which are almost always not desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,21 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- **[BREAKING]** The following folders are now excluded by default from both the
+  `files` and `output` arrays:
+
+  - `.git/`
+  - `.hg/`
+  - `.svn/`
+  - `.wireit/`
+  - `CVS/`
+  - `node_modules/`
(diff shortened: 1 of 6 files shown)


aomarks:
Oops, looks like there is an unexpected error happening. Will re-add reviewers when fixed.

aomarks:
> Oops, looks like there is an unexpected error happening. Will re-add reviewers when fixed.

Fixed!

augustjk (approved):
Great addition!

PR #549, opened 2022-11-19
Fix fast-glob issue with trailing slashes on ignored folders

aomarks:
I noticed a bug where if you pass a folder with a trailing slash to fast-glob's `ignore` list, it has no effect (e.g. you have to pass `"foo/ignore"`, not `"foo/ignore/"`)! Now we normalize exclude paths to account for this.


augustjk (approved):
Great find and fix!

justinfagnani (approved):

Issue #550, opened 2022-11-19
Account for fast-glob vs chokidar glob differences


aomarks:
fast-glob and chokidar understand globs in slightly different ways, plus we currently do a bunch of extra normalization etc. to the globs we pass to fast-glob, which we don't do to chokidar.The behaviors should be pretty much the same.

It's actually OK if chokidar sometimes handles more files than necessary, since we only use chokidar to notice a change *anywhere*, and always rely on fast-glob to figure out what *actually* changed for a given iteration (which could end up being nothing). Could indicate a performance issue, though, if we're watching more than we need to.

It's not OK if chokidar sometimes misses files that fast-glob does, since that means we won't trigger an iteration in all cases when we should.

PR #551, opened 2022-11-19
Run all glob tests against chokidar too

aomarks:
This PR runs every glob test against chokidar, since we should have identical behavior in theory. We currently don't, so there are a lot of tests that get excluded for the chokidar case.

Tracking the task of making them equivalent at https://github.com/google/wireit/issues/550, but figured I'd land the tests first so that we at least don't regress, and have something easy to track for improvements in future.

### DIFF
--- a/src/test/glob.test.ts
+++ b/src/test/glob.test.ts
@@ -120,491 +120,496 @@ test.after.each(async (ctx) => {
   }
 });
 
-test('empty patterns', ({check}) =>
-  check({
-    files: ['foo'],
-    patterns: [],
-    expected: [],
-  }));
+for (const mode of ['once', 'watch'] as const) {
+  test('empty patterns', ({check}) =>
+    check({
+      files: ['foo'],
+      patterns: [],
+      expected: [],
+    }));
 
-test('normalizes trailing / in pattern', ({check}) =>
-  check({
-    files: ['foo'],
-    patterns: ['foo/'],
-    expected: ['foo'],
-  }));
+  test('normalizes trailing / in pattern', ({check}) =>
+    check({
+      files: ['foo'],
+      patterns: ['foo/'],
+      expected: ['foo'],
+    }));
 
-test('normalizes ../ in pattern', ({check}) =>
-  check({
-    files: ['foo'],
-    patterns: ['bar/../foo'],
-    expected: ['foo'],
-  }));
+  test('normalizes ../ in pattern', ({check}) =>
+    check({
+      files: ['foo'],
+      patterns: ['bar/../foo'],
+      expected: ['foo'],
+    }));
 
-test('explicit file that does not exist', ({check}) =>
-  check({
-    files: [],
-    patterns: ['foo'],
-    expected: [],
-  }));
+  test('explicit file that does not exist', ({check}) =>
+    check({
+      files: [],
+      patterns: ['foo'],
+      expected: [],
+    }));
 
-test('* star', ({check}) =>
-  check({
-    files: ['foo', 'bar'],
-    patterns: ['*'],
-    expected: ['foo', 'bar'],
-  }));
+  test('* star', ({check}) =>
+    check({
+      files: ['foo', 'bar'],
+      patterns: ['*'],
+      expected: ['foo', 'bar'],
+    }));
 
-test('* star with ! negation', ({check}) =>
-  check({
-    files: ['foo', 'bar', 'baz'],
-    patterns: ['*', '!bar'],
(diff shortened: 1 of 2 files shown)


src/test/glob.test.ts
```diff
@@ -76,29 +80,60 @@ test.before.each(async (ctx) => {
...
+      } else if (mode === 'watch') {
+        const actual: string[] = [];
+        if (patterns.length > 0) {
+          const {watcher} = makeWatcher(
+            patterns,
+            rig.resolve(cwd),
+            () => undefined,
+            // We need ignoreInitial=false because we need the initial "add"
+            // events to find out what chokidar has found (we usually only care
+            // about changes, not intial files).
```
AndrewJakubowicz, line 118: Nit: `intial` -> `initial`.

src/test/glob.test.ts
```diff
@@ -120,491 +155,584 @@ test.after.each(async (ctx) => {
...
+      ],
+      patterns: ['symlink'],
+      expected: ['symlink'],
+    }));
+
+  test(`[${mode}] explicit directory excluded when includeDirectories=false`, ({
+    check,
+  }) => check({mode, files: ['foo/'], patterns: ['foo'], expected: []}));
+
+  if (mode !== 'watch') {
```
AndrewJakubowicz, line 254: Nit: Could abstract this check to be more readable. E.g. `skipWatchTest()', which could have a comment.
Optional.

src/test/glob.test.ts
```diff
@@ -76,29 +80,60 @@ test.before.each(async (ctx) => {
...
+        } else if (error !== undefined) {
+          throw error;
+        } else if (actual === undefined) {
+          throw new Error('Actual was undefined');
+        } else {
+          const actualPaths = actual.map((file) => file.path);
+          assert.equal(actualPaths.sort(), expected.sort());
+        }
+      } else if (mode === 'watch') {
+        const actual: string[] = [];
```
AndrewJakubowicz, line 110: Nit: Could add a symmetrical comment here with a similar sentiment to "`// One-time globbing (via fast-glob).`.

AndrewJakubowicz (approved):
Nice! Non blocking nits.

src/test/glob.test.ts
aomarks: Removed the other comment, since it's pretty much self-describing

src/test/glob.test.ts
aomarks: Done!

src/test/glob.test.ts
aomarks: Done

aomarks:
Sorry for the churn, I just force-pushed a much more incremental version of this PR that is easier to read.

Also switched to using the `test.skip` function, so that it's more obvious in the output how many are being skipped.

src/test/glob.test.ts
```diff
@@ -120,491 +154,646 @@ test.after.each(async (ctx) => {
...
-    followSymlinks: true,
-    includeDirectories: true,
-    expandDirectories: false,
-    throwIfOutsideCwd: false,
+for (const mode of ['once', 'watch'] as const) {
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  const skipIfWatch = mode === 'watch' ? test.skip : test;
+  const skipIfWatchOnWindows =
+    // eslint-disable-next-line @typescript-eslint/unbound-method
+    mode === 'watch' && IS_WINDOWS ? test.skip : test;
```
AndrewJakubowicz, line 162: Nice!

src/watcher.ts
```diff
@@ -448,22 +448,30 @@ const watchPathsEqual = (
...
+ * @param ignoreInitial Ignore the initial "add" events emitted when chokidar
+ * first discovers each file. We already do an initial run, so these events are
+ * just noise that may trigger an unnecessary second run.
+ * https://github.com/paulmillr/chokidar#path-filtering
+ */
+export const makeWatcher = (
   patterns: string[],
   cwd: string,
-  callback: () => void
+  callback: () => void,
```
AndrewJakubowicz, line 462: This is unrelated, but what does this callback do?

AndrewJakubowicz (approved):
Looks great!

src/test/glob.test.ts
AndrewJakubowicz: This `test.skip : test` technique is great!

src/watcher.ts
aomarks: It just gets passed to the `all` callback

PR #564, opened 2022-11-23
Improvements for yarn/pnpm and argument handling in general

aomarks:
- The `.yarn/` folder is now included in the default excludes, which is what yarn 3.x uses instead of `node_modules/`

- It's now allowed to use something like `../node_modules/.bin/wireit` as your script, instead of simply `wireit`. This is helpful when using yarn or pnpm, since neither of those automatically add parent directory `node_modules/` folders to `$PATH` in the way that npm does.

- We now automatically use `yarn.lock` or `pnpm-lock.yaml` instead of `package-lock.json` when we detect that you are using yarn or pnpm respectively.

- Slightly improved handling for chained (aka recursive) script invocations when using yarn 1.x. Due to https://github.com/yarnpkg/yarn/issues/8905, yarn 1.x acts differently to npm 6.x in the way that it sets its `npm_config_argv` variable. Instead of that variable reflecting the current script, it reflects the *first* script in a chain of scripts. That makes it difficult for us to reconstruct exactly which arguments were passed to the current script. We can now handle the case of `yarn run build --watch` (which is one of the main reasons you would have a recursive script), but still don't handle e.g. `yarn run build --watch -- --extra`, when invoked recursively (we don't notice the `--watch`). We at least can detect when we might not be understanding arguments correctly, and show a warning telling the user they should upgrade to yarn 3.x or switch to npm.

- Removed special handling for the `watch` argument. Since 0.6.0, a command like "npm run <wireitscript> watch" exited with an error, to help remind people to migrate to `--watch` instead. Enough time has elapsed now that we can just treat `watch` like any other argument, though.

### DIFF
--- a/src/cli-options.ts
+++ b/src/cli-options.ts
@@ -254,7 +254,7 @@ function getArgvOptions(
 /**
  * Try to find the npm user agent being used. If we can't detect it, assume npm.
  */
-function getNpmUserAgent(): 'npm' | 'yarnClassic' | 'yarnBerry' | 'pnpm' {
+function getNpmUserAgent(): Agent {
   const userAgent = process.env['npm_config_user_agent'];
   if (userAgent !== undefined) {
     const match = userAgent.match(/^(npm|yarn|pnpm)\//);


src/analyzer.ts
```diff
@@ -363,7 +363,11 @@ export class Analyzer {
     if (
       wireitConfig !== undefined &&
       scriptCommand.value !== 'wireit' &&
-      scriptCommand.value !== 'yarn run -TB wireit'
+      scriptCommand.value !== 'yarn run -TB wireit' &&
+      // This form is useful when using package managers like yarn or pnpm which
+      // do not automatically add all parent directory `node_modules/.bin`
+      // folders to PATH.
+      !/^(..\/)+node_modules\/\.bin\/wireit$/.test(scriptCommand.value)
```
justinfagnani: Is there any test for this?

src/cli-options.ts
```diff
@@ -194,14 +197,16 @@ export const getOptions = (): Result<Options> => {
     };
   }
 
+  const agent = getNpmUserAgent();
```
justinfagnani, line 184: Is there any warning for an unknown agent?

src/cli-options.ts
```diff
@@ -181,22 +181,6 @@ export const getOptions = (): Result<Options> => {
     return failureModeResult;
   }
 
-  if (process.argv[2] === 'watch') {
-    // TODO(aomarks) Remove in upcoming release.
```
justinfagnani, line 182: +1 nice deprecation

justinfagnani (approved):

src/cli-options.ts
augustjk: There is!
https://github.com/google/wireit/blob/3ed29fc16c642ecec85546284869156c44f0d803/src/cli-options.ts#L277-L281

Minor nit relating to that, the return type of the `getNpmUserAgent` can now use the `Agent` type.
https://github.com/google/wireit/blob/3ed29fc16c642ecec85546284869156c44f0d803/src/cli-options.ts#L265

augustjk (approved):

src/analyzer.ts
aomarks: Yep, see https://github.com/google/wireit/blob/1ff4a0ea89944447d8bcb99eb4916313095004bc/src/test/freshness.test.ts#L1527

src/cli-options.ts
aomarks: Done!

Issue #93, opened 2022-04-09
Ability to specify which environment variables should affect cache key


aomarks:

aomarks:
Proposal for syntax:

```
"wireit": {
  "build": {
    "env": {
      "foo": "bar",
      "baz": {
        "external": true
      }
    }
  }
}
```

So foo will be set to bar for that script, and baz will be just added to the fingerprint, with the value being inherited  (as it already would be).

justinfagnani:
btw, I found a nice workaround for my case. I put all my environment variables into a `.env` file that's read with the `dotenv` package. Then I depend on that in my files array.

It doesn't differentiate between env variable, so any change to the file will cause all scripts in the current graph that depend on it to re-run, but for me that works out well.

aomarks:
This feature was released in 0.9.1: https://github.com/google/wireit#environment-variables

Issue #580, opened 2022-12-02
Feature request: Allow services to be shared across wireit invocations


justinfagnani:
In some projects I might have many entrypoints into my script graph that share a common service. I might want to invoke these entrypoints separately while having them share that single service.

An example is the Firebase emulator. Generally a project has s single emulator config which supplies it with settings including ports to use. Starting the emulator multiple times isn't possible because of the fixed port. It's also a stateful service, so even if you solve the port issue you might want entrypoints to share the same state.

The entrypoints might be an app server, a separate dev server for component stories, and test entrypoints, which all need access to the emulator. I want to be able to run tests while the app server is running, or run app and stories servers.

Issue #556, opened 2022-11-21
How to handle cross-cutting environment variables


justinfagnani:
I want to change my builds slightly based on an environment variable, like `ELEVENTY_ENV` which would be valued `"development"` or `"production"`.

I have scripts in my configs that can change behavior based on env variables. I'd like to have two npm scripts to make invoking these easier. One plain script, and one script that sets the env variable first.

For a single script, I could try to do something like:

```json
"scripts": {
  "build:eleventy": {
    "command": "eleventy",
    ...
  },
  "build:eleventy:dev": {
    "command": "ELEVENTY_ENV=development npm run build:eleventy",
  },
}
```

but this becomes tedious when there are other scripts that call these. Each that does would need a `:dev` variant that depends on `:dev` variants itself.

I tried just setting the environment variable before calling `npm`, like `ELEVENTY_ENV=development npm run build`, except that wireit doesn't know that a script depends on an env var, so it won't rerun a script that appears clean, but actually changes output based on the variable.

Is there a better pattern for this?

aomarks:
https://github.com/google/wireit/issues/93 should address this

justinfagnani:
Ah, sorry I missed that!

PR #581, opened 2022-12-03
Add environment variable configuration

aomarks:
Adds the ability to either set environment variables directly, or to indicate that an externally-provided environment variable affects the behavior of the script.

Setting variables directly is useful because there is no built-in way to do this in a cross-platform way (especially macOS/Linux shells vs Windows shells). A typical solution is to use https://github.com/kentcdodds/cross-env (which is deprecated), but now wireit can take over this responsibility.

Indicating that an external environment variable affects the behavior of a script is useful because the value will be included in the fingerprint, meaning output will not be re-used if that value changed.

Example:

```json
{
  "wireit": {
    "bundle:prod": {
      "command": "rollup -c",
      "files": ["lib/**/*.js", "rollup.config.js"],
      "output": ["dist/bundle.js"],
      "env": {
        "MODE": "prod",
        "DEBUG": {
          "external": true
        }
      }
    }
  }
}
```

Fixes https://github.com/google/wireit/issues/93

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -524,6 +524,8 @@ export class Analyzer {
     );
     this._processPackageLocks(placeholder, packageJson, syntaxInfo, files);
 
+    const env = this._processEnv(placeholder, packageJson, syntaxInfo, command);
+
     // It's important to in-place update the placeholder object, instead of
     // creating a new object, because other configs may be referencing this
     // exact object in their dependencies.
@@ -542,6 +544,7 @@ export class Analyzer {
       configAstNode: wireitConfig,
       declaringFile: packageJson.jsonFile,
       services: [],
+      env,
     };
     Object.assign(placeholder, remainingConfig);
   }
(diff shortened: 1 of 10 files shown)


src/analyzer.ts
```diff
@@ -1180,6 +1183,115 @@ export class Analyzer {
...
+    command: JsonAstNode<string> | undefined
+  ): Record<string, string> {
+    if (syntaxInfo.wireitConfigNode === undefined) {
+      return {};
+    }
+    const envNode = findNodeAtLocation(syntaxInfo.wireitConfigNode, ['env']);
+    if (envNode === undefined) {
+      return {};
+    }
+    if (command === undefined) {
```
justinfagnani, line 1199: Does this mean passthrough commands can't set env vars?

What about a pair of `start` and `start:dev` scripts that are both pass-throughs where the only difference is that `start:dev` sets something like `MODE: "development"` for other scripts?

aomarks: > Does this mean passthrough commands can't set env vars?

Yep. I general `env` only applies to that specific script, it doesn't apply transitively, so it doesn't make sense if there is no `command`, since it would have no effect. (The non-transitiveness is documented in the README).

> What about a pair of `start` and `start:dev` scripts that are both pass-throughs where the only difference is that `start:dev` sets something like `MODE: "development"` for other scripts?

In that case I'd create two different branches of the dependencies. You'd probably want different output folders anyway.

justinfagnani: Hmm, maybe I should reopen https://github.com/google/wireit/issues/556

Because my question there is about how to avoid having a big duplication of scripts to handle the cross-cutting prod/dev cases. If I have separate scripts, then I don't have any need for an env variable to be an input to the fingerprint - I would either not use env vars, or each script would only ever be invoked with a single value of the variable.

justinfagnani: Another case where sharing env vars would be important is if we figure out how to start services with dynamic resources - like a server with a dynamic port. In that case we'd like to somehow extract the port and give it to anything that depends on it. That would allow for starting services multiple times (the opposite of what I'm asking for in https://github.com/google/wireit/issues/580, I think both behaviors would be useful).

This is two different directions of sharing: from dependent->dependency, and from dependency->dependent.

Taking just the dependent->dependency which I'm asking about originally, could we have a way to explicitly share the value via a property in dependencies?

```json
{
  "build" {
    "command": "...",
    "env": {
      "MODE": "external"
    }
    "files": [],
    "output": [],
  },
  "start": {
    "command": "...",
    "env": {
      "MODE": {
        "external": true,
    },
    "dependencies": {
      "build": {
        "env": {
          "MODE": {
            // pass "start"s MODE value to "build"
            "inherit": true,
          }
        }
      }
    }
  }
}
```

augustjk (approved):
Awesome!

src/analyzer.ts
augustjk: Then any dependency of the dependency would also need the env option too if you need it to propagate. Would it be a bad idea to add something like "cascade" at the script's env config, where it could be like this?
```json
{
  "start": {
    "command": "...",
    "env": {
      "MODE": {
        "value": "prod",
        "cascade": true
      }
    }
  }
}
```

aomarks: If an environment variable applies to all dependencies, then that implies that that environment variable should be included in the fingerprint. But very often that's not desirable, because it would mean you can't re-use output even though the variable doesn't actually matter for some particular script.

For example, if you had a dev and prod rollup script which differ via an environment variable., but both of them consumed the same `tsc` output, then you'd lose the cache hit for the `tsc` step.

---

Things also get a bit confusing when you have multiple paths to the same dependency, e.g. imagine this graph:

```
   A
  / \
 v   v
 B   C [BUILD=prod]
  \ /
   v
   D
```

So now D as traversed via B has BUILD=undefined, but as traversed via C has BUILD=prod. It's not clear to me what we should do there, and I suspect the answer is we shouldn't get ourselves in that situation to begin with.

justinfagnani (approved):

src/analyzer.ts
aomarks: Discussed some more offline with Justin -- I had actually missed the comment about setting `"inherit": true`, which addresses my concerns above pretty much. We also extended the idea to consider how you could directly reference the _specific_ script where you want to inherit an environment variable from. Also agreed that this can be a new extension to this feature, so we can land this initial support now.

PR #589, opened 2022-12-06
Bump wireit version and remove cross-env

aomarks:
Upgrades wireit to 0.9.1 and replaces use of `cross-env` with the new `env` setting.


AndrewJakubowicz (approved):
Nice!

Issue #568, opened 2022-11-26
Some services not killed correctly


justinfagnani:
I'm trying to figure out the conditions that cause this, and I don't have a repro yet.

I think it's a combination of:

- Service A as a dependency, like of a pass-through script
- Watch mode
- An error in a dependency of service A, like a TS build error for the service
- Fixing that error

Then if seems like when wireit goes to restart service A, it doesn't kill it first. This is end-user noticeable with ports still being in use, etc. At this point it becomes impossible to kill wireit with `^c`. You have to kill both `wireit` and the wayward service manually.

SanderElias:
@justinfagnani this seems to be the same problem I have in my angular project. Let me see if I can quickly whip up a demo repo.

deebloo:
I am seeing the situation when working with fastify and typescript. pretty frequently finding services still bound having not been terminated. I am experimenting with listening for process events and trying to close the server manually but I don't think that will actually solve it

deebloo:
a small reproducible example. tsc is set as a dependency of the server. Everything works perfectly until typescript throws an error. Once the error is fixed the previous service doesn't seem to have exited. 

It kind of seems like the execution of the graph STOPS when it hits and error in the tree (reasonable) but then the previous service is never cleaned up.

https://github.com/deebloo/wireit-service-error-test

aomarks:
Fix released in 0.9.2

deebloo:
thank you!

aomarks:
Thank you very much for the repro @deebloo !

Issue #540, opened 2022-11-15
WireIt doesn't kill `ng serve` properly


SanderElias:
This the follow-up on the discussion in #33 

I think your guess is right, and the problem is that WireIt doesn't seem able to kill `ng serve` properly.
If you look at this screenshot:
![image](https://user-images.githubusercontent.com/1249083/201902366-ca1b9d76-2d2d-46c5-bdc4-19f93eb759cd.png)

This happened after I killed WireIt Manually, and restarted the watch mode.
I `^c`'d it, and use `pkill` to kill the `ng serve` command. That resulted in 2 lines that seems to originate from WireIt.

SanderElias:
@aomarks Still happening. 
Still feel it would be better when there is an option to kill the service (and wait for the exit) before starting to build the dependencies.

SanderElias:
As promised in #568 I created a small repo to demo the issue I have.
steps to reproduce:

```bash
git clone git@github.com:SanderElias/wireiProblemRepo.git
cd wireiProblemRepo
npm i
npm run start
```

That should bring up the `ng serve` of the demo app. you can open this on `http://localhost:4200/`

Now open the file `projects/lib-a/src/lib/lib-a.service.ts` and change the string or something.
In most cases, the `ng serve` command will _not_ be killed at all. 
When you repeat this a couple of times, you probably will notice a build error from ng-serve that seems to stem from out-of-order execution. 

It would help to have the option to kill `ng serve` before the libs get compiled, but that is a separate issue.

PR #596, opened 2022-12-08
Fix service behavior after dependency failure in watch mode

aomarks:
Previously, if a failure occurred in a persistent (i.e. top-level) service's dependencies, and if we were in watch mode on the 2nd or later iteration, then we completely forgot about the previously running service.

This had two similar effects, both of which could manifest as e.g. "port not available" errors:

- If the dependency was eventually fixed, we would try to start the service again, even though the previous version was still running.

- If the user exited wireit with Ctrl-C, we would not shut down the child process because we lost track of it.

This PR fixes these problems by transitioning to a new "started-broken" state when a failure in an already-running service's dependency occurs, which keeps a reference to the child process.

This PR also fixes some failure logging issues, where we would sometimes log the same failure multiple times, or sometimes log a failure at the very end of execution instead of as soon as it actually happens.

Fixes https://github.com/google/wireit/issues/568
Fixes https://github.com/google/wireit/issues/540

cc @SanderElias @deebloo @justinfagnani If you'd like to test out this fix early, I've published a pre-release as `wireit@0.9.2-pre.1`. If you try it, let me know how it goes!


justinfagnani (approved):

deebloo:
@aomarks initial tests look good!

SanderElias:
@aomarks This seems to work well. (20 mins into testing. I'll report back if it starts failing later on, but I don't expect so)
However, I noticed that services are now actually watching the things in the files array and get restarted when changes are there. This has changed from before when services didn't do this. I assume it is by intention?

SanderElias:
@aomarks It doesn't "crash" my angular services anymore. 
However, I still run into an issue. Technically not a WireIt thing, but it is not solvable without a change in it, I'm afraid.
Take an condig like this:

```json
{
  "serve"{
    "command": "ng serve --open --configuration=dev",
    "files": [
      "libs/**"
    ],
    "service": true,
    "dependencies": [
      "build:libs"
    ]
  },
  "build:libs" : {
    "dependencies": [
      "build:lib1",
      "build:lib2"
    ]
  },
  "build:lib1": {
    "command": "ng build --prod --configuration=prod",
    "files": [
      "projects/libs/lib1/**"
    ],
    "output": [
      "libs/lib1"
    ]
  },
  "build:lib2": {
    "command": "ng build --prod --configuration=prod",
    "files": [
      "projects/libs/lib2/**"
    ],
    "output": [
      "libs/lib2"
    ]
  }
}
```

When one of the libraries fails to build, the 'libs' folder is updated with a faulty lib. When this happens, the `ng serve` will fail (as expected) but generates a boatload of output, obfuscating  the error in the dependency, 
I think the solution would be to have an option to 'stop' the service while the deps are rebuilding. Setting the cascade options doesn't seem to have an effect.

(BTW, fixing the error and saving a couple of times again will end up with everything going back to "normal")

Issue #606, opened 2022-12-15
Watchers can end up watching entire C:/ drive on windows. 


filimon-danopoulos-stratsys:
In some configurations it is possible for chokidar to end up watching the entire C:/ drive on windows machines. This in turn leads exceptions being thrown due to locked files being read.  
See the following reproduction repo https://github.com/filimon-danopoulos-stratsys/wireit-watch-issue for steps to reproduce. Note that the reproduction it is not 100% reliable, sometimes it takes a while for the condition to show. 

![](https://github.com/filimon-danopoulos-stratsys/wireit-watch-issue/blob/main/screen-grab.png?raw=true)

aomarks:
This should be fixed in `0.9.3`, which was just released. Let me know if you still encounter issues!

PR #628, opened 2023-01-03
Only register watchers for package-lock.json files that exist

aomarks:
### Background

We automatically add `package-lock.json` (or equivalent depending on the package manager) entries to the `files` of every script during analysis. We do this so that updates to dependencies invalidate the fingerprint and trigger re-runs.

However, previously we added all *potential* `package-lock.json` paths to the `files` array, whether or not they exist. Meaning for a script defined in `/foo/bar/package.json` we would add:
  - `/foo/bar/package-lock.json`
  - `/foo/package-lock.json`
  - `/package-lock.json`

This is technically correct, however the effect of this in watch mode is that we would register a chokidar watcher for all of those paths, which meant asking the OS to listen for any new file created in any of those directories (including the root directory).

It seems that on Windows, this will occasionally cause errors such as `EBUSY: resource busy or locked, lstat 'C:\DumpStack.log.tmp`. I'm a bit unclear as to why an `lstat` is being issued for a file in `C:\`, since it doesn't seem necessary to stat files that we aren't actually interested in -- but that is what happens through chokidar for some reason.

### This PR

We now only add `package-lock.json` files to the `files` array to a script if the `package-lock.json` exists at the time of analysis. This does mean that a new watch iteration won't occur if a new `package-lock.json` is created during a watch session, but that's really not very important, and not worth the cost/bugs associated with always watching all parent directory listings for all scripts.

Fixes https://github.com/google/wireit/issues/606

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Fixed
+
+- In watch mode, watchers are no longer created for `package-lock.json` files
+  that don't yet exist at the time of analysis. This saves resources, and on
+  Windows should reduce errors such as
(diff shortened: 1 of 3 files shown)


augustjk (approved):
Good find! Agreed that cost/benefit of handling creation of a new package-locks doesn't justify the extra overhead.

Issue #644, opened 2023-01-18
Don't require dependencies to be in the scripts section


aomarks:
We're quite strict about every dependency being a real npm script, but sometimes it's nice to have intermediate scripts that are purely defined in the wireit section, even if they can't be invoked directly. This should be allowed.

filimon-danopoulos-stratsys:
This might be a bit backwards in relation to this issue but I believe it is related. 

Assuming I can have "pure" wireit scripts I would like the ability to run the same scripts too but without having to expose them in `scripts`. That way my `package.json` scripts section can be kept very high level, `build`, `lint`, `test` etc without having to care about specifics and dependencies between scripts. 

```json
{ 
  "scripts": {
    "build": "yarn wireit:build",
  },
  "wireit": {
    "wireit:build": {
      "command": "esbuild src/index.ts",
      "clean": "if-file-deleted",
      "files": [
        "./src/**/*"
      ],
      "output": [
        "./dist/**"
      ],
      "dependencies": [
        {
          "script": "wireit:tsc",
          "cascade": false
        }
      ]
    },
    "wireit:tsc": {
      "command": "tsc --project tsconfig.json",
      "clean": true,
      "files": [
        "./src/**/*",
        "tsconfig.json"
      ],
      "output": [
        "./typings/**/*"
      ]
    }
  },
}
```

However when need I would like to tap in to the wireit scripts too, `wireit run wireit:tsc`. In my actual setup I run `lessc`, `esbuild`, `tsc`, and `webpack` as part of the build.

PR #653, opened 2023-01-29
Allow wireit-only scripts

aomarks:
Allows defining a script in the `wireit` section even if it doesn't have an entry in `scripts`. Such scripts can be used as dependencies of other wireit scripts, but can't be run directly.

This is nice for intermediate scripts that are created only for internal organization, and would not really make sense to invoke directly.

Fixes https://github.com/google/wireit/issues/644

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -456,7 +456,9 @@ export class Analyzer {
           reason: 'script-not-wireit',
           script: placeholder,
           diagnostic: {
-            message: `This command should just be "wireit", as this script is configured in the wireit section.`,
+            message:
+              `This command should just be "wireit", ` +
+              `as this script is configured in the wireit section.`,
             severity: 'warning',
             location: {
               file: packageJson.jsonFile,
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -14,7 +14,12 @@ import {Dependency, scriptReferenceToString, ServiceConfig} from './config.js';
 import {findNodeAtLocation, JsonFile} from './util/ast.js';
 import {IS_WINDOWS} from './util/windows.js';
 
-import type {ArrayNode, JsonAstNode, NamedAstNode} from './util/ast.js';
+import type {
+  ArrayNode,
+  JsonAstNode,
+  NamedAstNode,
+  ValueTypes,
+} from './util/ast.js';
 import type {Diagnostic, MessageLocation, Result} from './error.js';
 import type {Cycle, DependencyOnMissingPackageJson, Failure} from './event.js';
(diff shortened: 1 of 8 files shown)


src/analyzer.ts
```diff
@@ -117,6 +122,18 @@ const DEFAULT_LOCKFILES: Record<Agent, string[]> = {
...
 };
 
+function isValidWireitScriptCommand(command: string): boolean {
+  return (
+    command === 'wireit' ||
+    command === 'yarn run -TB wireit' ||
+    // This form is useful when using package managers like yarn or pnpm which
+    // do not automatically add all parent directory `node_modules/.bin`
+    // folders to PATH.
+    /^(\.\.\/)+node_modules\/\.bin\/wireit$/.test(command) ||
```
justinfagnani, line 132: Do you ever need to handle `./node_moduules/...`? Like in a non-monorepo with Yarn>

src/analyzer.ts
```diff
@@ -434,39 +415,89 @@ export class Analyzer {
         },
       });
     }
+    // It's important to in-place update the placeholder object, instead of
+    // creating a new object, because other configs may be referencing this
+    // exact object in their dependencies.
+    const remainingConfig: LocallyValidScriptConfig = {
+      ...placeholder,
```
justinfagnani, line 422: Why do you need this spread if you just assign back into `placeholder`?

src/analyzer.ts
```diff
@@ -434,39 +415,89 @@ export class Analyzer {
...
+      !isValidWireitScriptCommand(scriptCommand.value)
+    ) {
+      {
+        const configName = wireitConfig.name;
+        placeholder.failures.push({
+          type: 'failure',
+          reason: 'script-not-wireit',
+          script: placeholder,
+          diagnostic: {
+            message: `This command should just be "wireit", as this script is configured in the wireit section.`,
```
justinfagnani, line 459: nit: break up long string?

justinfagnani (approved):

src/analyzer.ts
aomarks: I think if it's installed to `./node_modules/.bin` then every package manager should work with just `wireit`, no need to write `./node_modules/...`. npm also automatically checks all parent directory `node_modules/.bin`, but yarn and pnpm don't do that. So this just handles that gap.

src/analyzer.ts
aomarks: So that the `:LocallyVaildScriptConfig` type annotation can be there, so that we know we're ending up with an object that fully satisfies that interface.

src/analyzer.ts
aomarks: Done

PR #658, opened 2023-01-30
Release 0.9.4

aomarks:


aomarks:
> looks like lockfile changed to v3 and some deps got updated. i think v3 lockfile isn't backwards compatible with node 14.

Oops, yeah that's a bit annoying. Will need to swap between npm versions until we can finally drop Node 14 support. Fixed.

augustjk (approved):

PR #664, opened 2023-02-03
Handle a closed socket with error as an error in github-actions-cache

rictic:
Hoping to catch ECONRESET errors this way

### DIFF
--- a/src/caching/github-actions-cache.ts
+++ b/src/caching/github-actions-cache.ts
@@ -630,6 +630,16 @@ function request(
       req.on('error', (error) => {
         resolve({ok: false, error});
       });
+      req.on('socket', (socket) => {
+        socket.on('error', () => {
+          resolve({ok: false, error: new Error('socket error')});
+        });
+        socket.on('close', (hadError) => {
+          if (hadError) {
+            resolve({ok: false, error: new Error('socket closed with error')});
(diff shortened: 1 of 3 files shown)


src/caching/github-actions-cache.ts
```diff
@@ -652,6 +652,11 @@ function readBody(res: http.IncomingMessage): Promise<string> {
     res.on('error', (error: Error) => {
       reject(error);
     });
+    res.socket.on('close', (hadError) => {
+      if (hadError) {
+        reject(new Error('socket closed with error'));
```
aomarks, line 657: I'm not sure this will be sufficient, since we don't currently try/catch the call to `readBody`. So I think we also need to add checks for this failure everywhere `readBody` is called, and then do the `_maybeHandleServiceDown` thing on failure. Converting this API to use `Result` instead of exceptions would also be cool.

rictic: Agree, but my goal is to first get the ECONNRESET errors to be in our stack trace rather than in the uncaught exceptions handler

aomarks (changes requested):
Nice! Changelog entry?

rictic:
Added!

aomarks (approved):

Issue #525, opened 2022-11-12
Display summary with timing data


aomarks:
On each iteration, show a summary of how many scripts ran/were fresh/were restored from cache -- and how long it took.

joshkraft:
Hi @aomarks - working on a [POC](https://github.com/google/wireit/compare/main...joshkraft:wireit:enhance-execution-logging?expand=1) of this enhancement and had a quick question. 

Do you think it would be appropriate to implement this within the `default-logger.ts`, or would it be better for that to stay relatively stateless? As an alternative, it seems like it might be possible to implement this in `standard.ts`.

aomarks:
Hi @joshkraft. Cool! Definitely welcome a contribution here!

Probably many valid ways to do this, but my first thought would be a new class called e.g. `SummaryLogger` which takes another logger to pass events through to, but as it sees events goes by, also keeps a counter of the interesting events we want to summarize, and then when you call `.printSummary()` or something it displays that and maybe resets the counters? That way we could call it once when we're in normal mode, and on each iteration when we're in watch mode. 

Also note I *just* added a new `WatchLogger` at https://github.com/google/wireit/blob/main/src/logging/watch-logger.ts -- I think we'd want to instantiate the `SummaryLogger` both here for watch mode, and also in `cli.ts` for standard mode. (I'm assuming we want summaries for watch mode, but that's actually a good question -- do we?)

It would also be great to discuss here exactly what counts we should actually display -- have you thought about what you'd find most useful?

joshkraft:
Nice, I like the SummaryLogger approach. I'll give that a shot.

As far as logging in watch mode - I'd personally be fine with having the summary display in both watch mode and standard mode. Taking a look at `WatchLogger`, it seems like it might make sense to only display the summary on 'interesting' iterations?

For stats - so far I've got something like this:

```
🏁 [summary] Executed 9 script(s) in 1.21 seconds
    Ran:              9 (100%)
    Skipped (fresh):  0 (0%)
    Skipped (cached): 0 (0%)
```

I could see an argument for making this more verbose (error stats?), or more concise (does the user care if something was skipped because it was fresh or cached?). 

Unfortunately, I haven't used wireit enough at this point to have an educated opinion on what would be most useful. Do you have any thoughts?

aomarks:
> Nice, I like the SummaryLogger approach. I'll give that a shot.
> 
[...]

PR #691, opened 2023-02-21
Implement logging of metrics

joshkraft:
Adds a MetricsLogger class that keeps track of several metrics related to the execution of scripts:

- total successes (ignoring `no-command`'s, which don't run anything)
- scripts that ran
- scripts that were fresh 
- scripts that were restored from cache
- elapsed time

In 'normal' mode, metics are logged on each iteration with at least one 'success'. In 'watch mode', metrics are logged on each _interesting_ iteration with at least one 'success'.

metrics

 Addresses issue #525

### DIFF
--- a/src/logging/default-logger.ts
+++ b/src/logging/default-logger.ts
@@ -306,4 +306,8 @@ export class DefaultLogger implements Logger {
       }
     }
   }
+
+  printMetrics(): void {
+    // printMetrics() not used in default-logger.
+  }
 }
--- a/src/logging/metrics-logger.ts
+++ b/src/logging/metrics-logger.ts
@@ -17,7 +17,9 @@ export class MetricsLogger implements Logger {
   private readonly _metrics: Metric[] = [
     {
       name: 'Success',
-      matches: (e: Event) => e.type === 'success',
+      // 'no-command' is technically a success, but we don't want to count it as
+      // a success for this metric because nothing was actually run.
+      matches: (e: Event) => e.type === 'success' && e.reason !== 'no-command',
       count: 0,
     },
     {
--- a/src/logging/metrics-logger.ts
+++ b/src/logging/metrics-logger.ts
@@ -1,3 +1,9 @@
+/**
+ * @license
+ * Copyright 2023 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 import {hrtime} from 'process';
 import {Event} from '../event.js';
 import {DefaultLogger} from './default-logger.js';
--- a/src/logging/metrics-logger.ts
+++ b/src/logging/metrics-logger.ts
@@ -31,7 +31,7 @@ export class MetricsLogger implements Logger {
       count: 0,
     },
     {
-      name: 'Skipped (cached)',
+      name: 'Restored from cache',
       matches: (e: Event) => e.type === 'success' && e.reason === 'cached',
       count: 0,
     },
@@ -61,7 +61,7 @@ export class MetricsLogger implements Logger {
     }
 
     const elapsed = this._getElapsedTime();
-    const nameOffset = 16;
+    const nameOffset = 20;
 
     const out: string[] = [
       `🏁 [metrics] Executed ${successes} script(s) in ${elapsed} seconds`,
--- a/src/logging/metrics-logger.ts
+++ b/src/logging/metrics-logger.ts
@@ -56,6 +56,11 @@ export class MetricsLogger extends DefaultLogger {
    * Update relevant metrics for an event and pass it up to the parent logger.
    */
   override log(event: Event): void {
+    // When in watch mode, metrics should reset at the start of each run.
+    if (event.type === 'info' && event.detail === 'watch-run-start') {
+      this._resetMetrics();
+    }
+
     this._updateMetrics(event);
     super.log(event);
   }
--- a/src/logging/metrics-logger.ts
+++ b/src/logging/metrics-logger.ts
@@ -1,6 +1,6 @@
 import {hrtime} from 'process';
 import {Event} from '../event.js';
-import {Logger} from './logger.js';
+import {DefaultLogger} from './default-logger.js';
 
 interface Metric {
   name: string;
@@ -11,8 +11,7 @@ interface Metric {
 /**
  * A {@link Logger} that keeps track of metrics.
  */
-export class MetricsLogger implements Logger {
-  private readonly _actualLogger: Logger;
+export class MetricsLogger extends DefaultLogger {
   private _startTime: [number, number] = hrtime();
   private readonly _metrics: Metric[] = [
     {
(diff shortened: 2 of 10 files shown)


src/logging/metrics-logger.ts
```diff
@@ -0,0 +1,112 @@
...
+      matches: (e: Event) => e.type === 'success',
+      count: 0,
+    },
+    {
+      name: 'Ran',
+      matches: (e: Event) => e.type === 'success' && e.reason === 'exit-zero',
+      count: 0,
+    },
+    {
+      name: 'Skipped (fresh)',
```
rictic, line 29: I don't think it'll be very clear from a user's perspective how`Skipped (cached)` is different from `Skipped (fresh)`. Consider instead `Restored from cache`

rictic (commented):
One high level comment: I think printMetrics should be a method on Logger

If we want a MetricsLogger that behaves like DefaultLogger only it prints metrics then we'd give DefaultLogger a printMetrics method that does nothing, and MetricsLogger would be a subclass that overrides it

src/logging/metrics-logger.ts
joshkraft: Agreed - changing to `Restored from cache`

joshkraft:
> One high level comment: I think printMetrics should be a method on Logger
> 
> If we want a MetricsLogger that behaves like DefaultLogger only it prints metrics then we'd give DefaultLogger a printMetrics method that does nothing, and MetricsLogger would be a subclass that overrides it

Hi @rictic - thanks for the feedback. I took your suggestion and refactored this a bit. Would you mind taking a look to make sure what I implemented aligns with that you had in mind? Thanks!

src/logging/default-logger.ts
```diff
@@ -306,4 +306,8 @@ export class DefaultLogger implements Logger {
       }
     }
   }
+
+  printMetrics(): void {
```
joshkraft, line 310: When this method is empty, eslint complains:

> Unexpected empty method 'printMetrics'.eslint@typescript-eslint/no-empty-function

This could also be resolved by adding `//eslint-disable-line` or updating the ESLint config, but adding a comment seemed pragmatic.

src/logging/watch-logger.ts
```diff
@@ -46,6 +48,10 @@ export class WatchLogger implements Logger {
     }
   }
 
+  printMetrics(): void {
```
joshkraft, line 51: When this method is empty, eslint complains:

> Unexpected empty method 'printMetrics'.eslint@typescript-eslint/no-empty-function

This could also be resolved by adding `//eslint-disable-line` or updating the ESLint config, but adding a comment seemed pragmatic.

src/logging/default-logger.ts
rictic: +1

rictic (approved):
Looks good to me!

src/test/metrics.test.ts
```diff
@@ -0,0 +1,323 @@
+import {suite} from 'uvu';
```
rictic, line 1: Please include a copyright header at the top of this file, with a 2023 copyright date

rictic (changes requested):
Ok, I've tested it locally, this works great! Just one small thing and it's good to go

src/logging/metrics-logger.ts
```diff
@@ -0,0 +1,117 @@
...
+  private _updateMetrics(event: Event): void {
+    for (const metric of this._metrics) {
+      if (metric.matches(event)) {
+        metric.count++;
+      }
+    }
+  }
+
+  private _resetMetrics(): void {
+    this._startTime = hrtime();
```
rictic, line 97: This does the wrong thing in watch mode I think. We end up reporting the time from the end of the last run, rather than the beginning of this run.

So e.g. if you go get coffee for five minutes, then come back and trigger a build that takes three seconds, the elapsed time will be reported as five minutes and three seconds

rictic (changes requested):

src/test/metrics.test.ts
joshkraft: Done!

src/logging/metrics-logger.ts
joshkraft: Nice catch. I just added some code to reset the metrics when a `watch-run-start` event is received, I assume this is when we would want to 'start the timer'?

src/logging/metrics-logger.ts
```diff
@@ -0,0 +1,128 @@
...
+  constructor(rootPackage: string) {
+    super(rootPackage);
+  }
+
+  /**
+   * Update relevant metrics for an event and pass it up to the parent logger.
+   */
+  override log(event: Event): void {
+    // When in watch mode, metrics should reset at the start of each run.
+    if (event.type === 'info' && event.detail === 'watch-run-start') {
```
aomarks, line 60: Not the biggest deal, but it seems a little asymmetrical that this class intrinsically understands when it should reset its metrics, but requires the caller to tell it when to display the metrics.

One idea I had was to make each instance of a `MetricsLogger` only live for the duration of a single iteration. So, in one-shot mode there would be one instance, but in watch mode we would create a new one on each iteration. The benefit of that is that we don't need any reset logic -- we can just use class field initializers to set the counters to 0 and the start time to now.

I'm also fine with it as-is, though.

src/logging/metrics-logger.ts
aomarks: Yes I think this is correct now :+1:

src/test/metrics.test.ts
aomarks: Seems like the copyright header is still missing from this file.

aomarks (changes requested):
Awesome work!

src/test/metrics.test.ts
joshkraft: Oops, just pushed up the header to this file.

src/logging/metrics-logger.ts
joshkraft: I think this is a good suggestion, looking into it. I explored this a bit when I got started with this work, but couldn't identify an existing event that signals 'the last service has finished executing', which I assume is when we would want MetricsLogger to automatically print metrics to the console. 

Do you know if there is an existing way to know when all services have finished executing, or would we need a new event (or some other signal)?

aomarks: Yeah, maybe it's not so straightforward as I imagined. This is great as-is, we can always refactor things later if needed.

aomarks (approved):

rictic (approved):

thiagowfx:
Drive-by: Is it possible to opt-out of metrics? Any flag or switch we can add to do so? (Apologies if this is already documented elsewhere, I could not find a way to do so.)

joshkraft:
> Drive-by: Is it possible to opt-out of metrics? Any flag or switch we can add to do so? (Apologies if this is already documented elsewhere, I could not find a way to do so.)

Hi @thiagowfx - this wasn't originally created with any opt-out mechanism. FWIW, the metrics are not aggregated or sent out - it stays local. Is your concern just around reducing the amount of output when executing commands?

thiagowfx:
> Is your concern just around reducing the amount of output when executing commands?

Yes, pretty much only that.

A typical `wireit` command in our repository () now looks like this with version `0.10.0`:
image

...and it's not particularly actionable.

thiagowfx:
With `--watch` it's even noisier, e.g.:

image

rehanvdm:
Ticket to disable/opt-out of metrics https://github.com/google/wireit/issues/830

PR #729, opened 2023-03-29
Update node runners

rictic:
16 is now the most recent maintenance release, 18 the LTS and 20 the current

This came up because pnpm v8 doesn't support node 14

The operating system options haven't changed yet: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,11 +20,11 @@ jobs:
             os: windows-2022
           - node: 18
             os: ubuntu-22.04
-          - node: 19
+          - node: 20
             os: ubuntu-22.04
-          - node: 19
+          - node: 20
             os: macos-12
-          - node: 19
+          - node: 20
             os: windows-2022
           - node: 20
             os: ubuntu-22.04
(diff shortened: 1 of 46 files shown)


aomarks:
I was thinking we'd end Node 14 support when Node 14 goes out of LTS, which is in 1 month

https://github.com/nodejs/Release#release-schedule

Then we could also update the docs here https://github.com/google/wireit#requirements, and we should also be able to ship native private fields I think?

rictic:
+1, delaying this change until then

aomarks (approved):
Nice!

WDYT about also changing the JS we emit so that this is *actually* breaking? We could bump up the ES level, and turn on native class fields. Could also migrate to #private class fields.

Otherwise when we eventually do rely on a newer feature, we might in practice need another breaking change, because people won't have realized that this change was the breaking one.

rictic:
Hm, github caching tests are locking up in CI with node 20. I don't have time to debug today though.

aomarks (approved):
Awesome!

We could now also migrate from uvu to the new built-in Node test runner, since that's available since 18.

PR #805, opened 2023-07-07
Update some deps

rictic:


aomarks (approved):

AndrewJakubowicz:
Looks good - although strange that Node 14 tests are broken.

rictic:
Our deps are breaking their node 14 support, as it's no longer supported, meanwhile a couple of our tests are broken in node 20. We need to dive in and investigate that

PR #836, opened 2023-08-26
Add a new, more quiet logger

rictic:
Example of it in action:

https://github.com/google/wireit/assets/1659/c1f9cc34-0340-4b52-9ac5-8c9f20552196

Also supports watch mode:

https://github.com/google/wireit/assets/1659/72f5d7e0-1d8e-46c4-9ba8-021d19c3df46

Supports services too, emitting their output immediately.

PR #838, opened 2023-08-30
Some small cleanup changes pulled out of the quiet logger PR

rictic:
#836 is way too big, this pulls some of the incidental improvements out of it so they can be reviewed in isolation

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,6 +2,6 @@
   "javascript.preferences.importModuleSpecifierEnding": "js",
   "typescript.preferences.importModuleSpecifierEnding": "js",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "typescript.tsdk": "node_modules\\typescript\\lib",
+  "typescript.tsdk": "node_modules/typescript/lib",
   "typescript.enablePromptUseWorkspaceTsdk": true
 }
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -29,6 +29,7 @@ import type {
   ScriptReferenceString,
 } from './config.js';
 import type {Agent} from './cli-options.js';
+import {Logger} from './logging/logger.js';
 
 export interface AnalyzeResult {
   config: Result<ScriptConfig, Failure[]>;
@@ -147,9 +148,11 @@ export class Analyzer {
   private readonly _ongoingWorkPromises: Array<Promise<undefined>> = [];
   private readonly _relevantConfigFilePaths = new Set<string>();
   private readonly _agent: Agent;
+  private readonly _logger: Logger | undefined;
 
-  constructor(agent: Agent, filesystem?: FileSystem) {
+  constructor(agent: Agent, logger?: Logger, filesystem?: FileSystem) {
     this._agent = agent;
+    this._logger = logger;
     this._packageJsonReader = new CachingPackageJsonReader(filesystem);
   }
 
@@ -198,6 +201,29 @@ export class Analyzer {
    * a cycle in the dependency graph.
    */
   async analyze(
+    root: ScriptReference,
+    extraArgs: string[] | undefined
+  ): Promise<AnalyzeResult> {
+    this._logger?.log({
+      type: 'info',
+      detail: 'analysis-started',
+      script: root,
+    });
+    const analyzeResult = await this._actuallyAnalyze(root, extraArgs);
+    let rootScriptConfig = undefined;
+    if (analyzeResult.config.ok) {
+      rootScriptConfig = analyzeResult.config.value;
+    }
+    this._logger?.log({
+      type: 'info',
+      detail: 'analysis-completed',
+      script: root,
+      rootScriptConfig,
+    });
+    return analyzeResult;
+  }
+
+  private async _actuallyAnalyze(
     root: ScriptReference,
     extraArgs: string[] | undefined
   ): Promise<AnalyzeResult> {
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -210,15 +210,13 @@ export class Analyzer {
       script: root,
     });
     const analyzeResult = await this._actuallyAnalyze(root, extraArgs);
-    let rootScriptConfig = undefined;
-    if (analyzeResult.config.ok) {
-      rootScriptConfig = analyzeResult.config.value;
-    }
     this._logger?.log({
       type: 'info',
       detail: 'analysis-completed',
       script: root,
-      rootScriptConfig,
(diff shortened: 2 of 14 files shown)


.vscode/settings.json
```diff
@@ -2,6 +2,6 @@
   "javascript.preferences.importModuleSpecifierEnding": "js",
   "typescript.preferences.importModuleSpecifierEnding": "js",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "typescript.tsdk": "node_modules\\typescript\\lib",
+  "typescript.tsdk": "node_modules/typescript/lib",
```
AndrewJakubowicz, line 5: Could this impact Windows users?

src/analyzer.ts
```diff
@@ -200,6 +203,29 @@ export class Analyzer {
...
+    this._logger?.log({
+      type: 'info',
+      detail: 'analysis-started',
+      script: root,
+    });
+    const analyzeResult = await this._actuallyAnalyze(root, extraArgs);
+    let rootScriptConfig = undefined;
+    if (analyzeResult.config.ok) {
+      rootScriptConfig = analyzeResult.config.value;
+    }
```
AndrewJakubowicz, line 216: This is very nice. What do you think about inlining this as a ternary in value position of `rootScriptConfig`? Then you can skip this if there isn't a logger.

Although it may be a bit less readable. Up to you.

src/event.ts
```diff
@@ -377,6 +380,23 @@ export interface OutputModified extends InfoBase {
...
+ * files, analyze their wireit configs, and build the dependency graph.
+ */
+export interface AnalysisStarted extends InfoBase {
+  detail: 'analysis-started';
+}
+
+/**
+ * The analysis phase for an execution has begun, where we load package.json
+ * files, analyze their wireit configs, and build the dependency graph.
+ */
```
AndrewJakubowicz, line 394: Update to reflect completion of analysis.

src/execution/service.ts
```diff
@@ -818,6 +827,11 @@ export class ServiceScriptExecution extends BaseExecutionWithCommand<ServiceScri
   private _onChildStarted() {
     switch (this._state.id) {
       case 'starting': {
+        this._logger.log({
+          script: this._config,
+          type: 'info',
+          detail: 'service-process-started',
```
AndrewJakubowicz, line 833: Nit: Is it worth reflecting the state names in `detail`? E.g., `service-process-starting`?

src/execution/service.ts
```diff
@@ -262,6 +262,15 @@ export class ServiceScriptExecution extends BaseExecutionWithCommand<ServiceScri
       entireExecutionAborted,
       adoptee,
     };
+    // Doing this here ensures that we always log when the
+    // service stops, no matter how that happens.
```
AndrewJakubowicz, line 266: Are there other conditions outside the `stopping` and `failing` states that could be missed otherwise?

I do like hanging this off the promise.

edit: is `child.kill({force: true});`  one of these additional cases that are now handled?

src/event.ts
```diff
@@ -31,7 +31,7 @@ interface EventBase<T extends PackageReference = ScriptReference> {
 // Success events
 // -------------------------------
 
-type Success = ExitZero | NoCommand | Fresh | Cached;
+export type Success = ExitZero | NoCommand | Fresh | Cached;
```
AndrewJakubowicz, line 34: Nit, when exporting should add a jsdoc comment.

src/logging/default-logger.ts
```diff
@@ -310,4 +293,38 @@ export class DefaultLogger implements Logger {
...
+    return new WatchLogger(this);
+  }
+}
+
+/**
+ * Make a concise label for a script, or for just a package if we don't know
+ * the script name. If the package is different to the root package, it is
+ * disambiguated with a relative path.
+ */
+export function labelForScript(
```
AndrewJakubowicz, line 307: deduplication yay! very nice!

src/test/util/test-rig.ts
```diff
@@ -318,7 +322,11 @@ class ExecResult {
       // own process group. Passing the negative of the child's pid kills all
       // processes in the group (without the negative only the leader "sh"
       // process would be killed).
-      process.kill(-this._child.pid, 'SIGINT');
+      if (force) {
+        process.kill(-this._child.pid, 9);
```
AndrewJakubowicz, line 326: Nit: thoughts about using `'SIGKILL'` instead of the number?

AndrewJakubowicz (approved):
Nice!

.vscode/settings.json
rictic: Often `/` works everywhere but `\` only works on Windows. If it does then we could likely remove this by then

src/analyzer.ts
rictic: Ah, I like htat!

src/event.ts
```diff
@@ -392,7 +412,15 @@ export interface WatchRunEnd extends InfoBase {
...
+export interface ServiceProcessStarted extends InfoBase {
+  detail: 'service-process-started';
+}
+
+/**
+ * A service started running and if it has a readyWhen condition,
+ * that condition is met.
  */
 export interface ServiceStarted extends InfoBase {
   detail: 'service-started';
```
aomarks, line 426: How about renaming this old event to `service-ready`, to distinguish it more clearly from `service-process-started`? The "ready" terminology is what we use in the configuration file and docs to mean that the ready condition, if there is one, has been satisfied (e.g. "Server started" was logged).

src/event.ts
rictic: Good catch! Done

aomarks (approved):

src/event.ts
rictic: +1, done. Also logged `[serviceName] Service ready` in the default logger instead of `[serviceName] Service started`

src/execution/service.ts
rictic: Before this change we'd only log that it stopped if we stopped it outselves and it was in certain specific states. This way, no matter how it dies (e.g. it chooses to call process.exit()), we inform the logger about it.

src/execution/service.ts
rictic: Starting makes sense to the internal state machine of ServiceScriptExecution, but for everywhere else at this point we have a child process now, so the process has started.

src/test/util/test-rig.ts
rictic: Yeah that seems better

src/event.ts
rictic: Done!

PR #850, opened 2023-08-31
Improve ci output

rictic:


src/cli-options.ts
```diff
@@ -194,6 +194,9 @@ export const getOptions = (): Result<Options> => {
     if (str === 'quiet') {
       return {ok: true, value: new QuietLogger(packageRoot)};
     }
+    if (str === 'quiet-ci') {
+      return {ok: true, value: new QuietCiLogger(packageRoot)};
+    }
```
augustjk, line 199: This can be addressed later when changing default to 'quiet' as well, but wdyt about defaulting to 'quiet-ci' based on `process.env['CI']`? We have precedence for that for the cache option here https://github.com/google/wireit/blob/9a78f7ff363fb9653afb9ca2177deaaa9581c2fb/src/cli-options.ts#L126-L142

src/test/quiet-logger.test.ts
```diff
@@ -0,0 +1,114 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
```
aomarks, line 3: 2023

src/logging/quiet/writeover-line.ts
```diff
@@ -13,78 +13,63 @@ import {DEBUG} from '../logger.js';
...
+  updateStatusLine(line: string): void;
+  clearUntilDisposed(): Disposable | undefined;
+}
 
+abstract class BaseWriteoverLine implements StatusLineWriter {
+  private _updateInterval: NodeJS.Timeout | undefined;
+  protected _line = '';
+  protected _targetFps = 60;
+  /**
+   * If true, we write over the previous line with a \r carraige return,
```
aomarks, line 27: carriage

aomarks (approved):

src/cli-options.ts
rictic: Yes, agree. When we default to quiet we should default to quiet-ci in ci.

src/test/quiet-logger.test.ts
rictic: Done

src/logging/quiet/writeover-line.ts
rictic: Done

PR #851, opened 2023-09-01
Conditionally polyfill `Symbol.dispose`

jonkoops:
Only assigns a value to `Symbol.dispose` if a polyfill is required. This prevents an exception in environments that support `Symbol.dispose` natively, which will throw when assigning a value (as this is a read-only property), even if that value is a reference to `Symbol.dispose` itself.

For example, without this change you will get the following exception running Node.js v20.5.1:

```log
file:///home/jon/Projects/keycloak/keycloak/js/node_modules/.pnpm/wireit@0.11.0/node_modules/wireit/lib/logging/quiet/writeover-line.js:9
    Symbol.dispose =
                   ^

TypeError: Cannot assign to read only property 'dispose' of function 'function Symbol() { [native code] }'
```


agagancarczyk (approved):

rictic (approved):
Ah, good catch!

PR #853, opened 2023-09-01
Use `using` sytax to reliably dispose of http sockets

rictic:
Locally this makes us compatible with node 20

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,7 +35,6 @@ jobs:
           - node: 20
             os: windows-2022
 
-
       # Allow all matrix configurations to complete, instead of cancelling as
       # soon as one fails. Useful because we often have different kinds of
       # failures depending on the OS.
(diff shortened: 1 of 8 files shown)


rictic:
This is a change that only adds CI tasks. I'd like to release v0.12.0 with quiet-ci and node 20 testing, then immediately release v0.13.0 that drops node 14.

.github/workflows/tests.yml
```diff
@@ -28,6 +28,12 @@ jobs:
             os: macos-12
           - node: 19
             os: windows-2022
+          - node: 20
```
augustjk, line 31: Now that we're adding 20, perhaps we can drop 14 and expand 16 to also include macos and windows?
I also question whether testing on 19 is necessary. I've always considered odd numbered node versions to be unstable.

rictic: Yes, updating runners in https://github.com/google/wireit/pull/729

I'd like to get one release out that supports 14-20 and then an immediate follow up that drops that to 16-20 (followed shortly by another that drops it to 18-20)

aomarks (approved):
Neat

PR #856, opened 2023-09-01
Switch the default logger to "quiet"

rictic:
Draft PR because we want to give the 'quiet' logger a week or so for people to try it out before releasing this.


rictic:
We've heard no complaints, I think this is solid, let's switch over and get a release out rather than blocking on explain and deflake

aomarks (approved):
Woohoo -- just a nit about the changelog date

PR #859, opened 2023-09-04
A few minor fixups

rictic:


src/test/ide.test.ts
```diff
@@ -565,4 +565,43 @@ test('can jump from colon in scripts section to wireit config', async ({
   });
 });
 
+test('can handle a script with  from colon in scripts section to wireit config', async ({
```
aomarks, line 568: typo in this description I think?

aomarks (changes requested):
Nice! using.

src/test/ide.test.ts
rictic: Wasn't meant to be part of this change, removed!

aomarks (approved):

Issue #803, opened 2023-06-30
Internal error: Supposedly valid service did not have command


leumasme:
A config like
```
    "foo": {
      "service": true,
      "dependencies": [
        "bar"
      ]
    },
```
is not highlighted as invalid by the extension.
Having such a config seems to cause the extension to crash repeatedly
```
Error: Internal error: Supposedly valid service did not have command
    at on._checkForCyclesAndSortDependencies (c:\Users\Temm\.vscode\extensions\google.wireit-0.6.0\server.js:42:20959)
    at on.analyzeFiles (c:\Users\Temm\.vscode\extensions\google.wireit-0.6.0\server.js:42:4600)
    at async ii.getDiagnostics (c:\Users\Temm\.vscode\extensions\google.wireit-0.6.0\server.js:42:24688)
    at async Au (c:\Users\Temm\.vscode\extensions\google.wireit-0.6.0\server.js:42:29636)
[Error - 2:30:09 AM] The wireit server server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.
```
It seems that the extension only checks wether any of `command`, `dependencies`, `files` are present, and not the (undocumented?) rule that a service must have a command
https://github.com/google/wireit/blob/28cb48bb13ee4d0800dd9c491b10f6ec3aa14fdc/src/analyzer.ts#L518-L541
https://github.com/google/wireit/blob/28cb48bb13ee4d0800dd9c491b10f6ec3aa14fdc/src/analyzer.ts#L1529-L1537
> we should never get here. We throw here to convince TypeScript that this is guaranteed.

Very amusing comment

PR #860, opened 2023-09-04
Don't upgrade a script to locally-valid if it has errors

rictic:
This way later scripts can assume locally-valid means no failures in the script that are locally determinable.

I think this was the original intent, and some code seems to assume this. See e.g. https://github.com/google/wireit/issues/803


aomarks (approved):
👍

aomarks (changes requested):
Oh, add CHANGELOG entry for this under "Fixed"?

rictic:
Done

aomarks (approved):

Issue #806, opened 2023-07-10
Introduce wireit 'explain'


kwangure:
Sometimes tracking why a particular command in a monorepo was run takes a while and hinges a lot on understanding deeply how Wireit works, which might not always be the case for everyone working with the codebase. Being able to run `wireit explain <command>` similar to `tsc --explainFiles` to see why a particular dependent command is included/cached would be helpful. It would make it a lot easier to refine `files`/`outputs` passed to Wireit.

I imagine it would by a dry-run type of command, since we often don't want to modify the cache/files we're trying to observe. This would be run directly rather than through `npm run`.

PR #863, opened 2023-09-05
Add a new "explain" logger

rictic:
A somewhat different design, but intended to address the use case brought up in https://github.com/google/wireit/issues/806 and that we've frequently needed ourselves.


aomarks (approved):
Looks great! Some pretty minor comments.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Added a new logger for debugging your Wireit configs. Set `WIREIT_LOGGER=explain` to see detailed information about why each script is run, as well as the normal `simple` logs, without any stdout or stderr of scripts.
```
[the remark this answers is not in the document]
rictic: Done

src/test/gc.test.ts
```diff
@@ -74,6 +74,7 @@ async function retryWithGcUntilCallbackDoesNotThrow(
   cb: () => void,
 ): Promise<void> {
   for (const wait of [0, 10, 100, 500, 1000]) {
+    collectGarbage();
```
aomarks, line 77: 1 GC not enough?

aomarks (approved):

src/test/gc.test.ts
rictic: Surprisingly, it often isn't. When doing memory leak tests in Lit we found that memory is often not freed unless you call gc twice in a row.

🤷

filimon-danopoulos:
@rictic @aomarks Is it possible to get this up to date and merged? I'm in the process of troubleshooting wireit cache invalidation and I am currently fumbling in the dark.  This could potentially help a lot.

PR #893, opened 2023-09-28
Update our `npx wireit` detection

rictic:
The latest version of npm doesn't set the `npm_execpath` env variable.


aomarks (approved):
Do we need to check both, for compatibility with older npm versions?

Also a test would be really nice because I'm pretty sure it will change again sometime :)

rictic:
> Also a test would be really nice

This is actually in response to our tests now failing consistently on node 20. Thanks past-Al! 

I manually tested that this works back to node 16, so I think this is a solid solution

aomarks:
> This is actually in response to our tests now failing consistently on node 20. Thanks past-Al!
> 
> I manually tested that this works back to node 16, so I think this is a solid solution

Oh, great!

PR #894, opened 2023-09-30
Downloading vscode for the tests is flaky, so retry a couple times.

rictic:


vscode-extension/src/test/scripts/runner.ts
```diff
@@ -8,13 +8,33 @@ import * as path from 'path';
...
+    try {
+      await run();
+      return;
+    } catch (err) {
+      console.error('Failed to run tests, retrying...');
+    }
+    // wait a few seconds before retrying
+    await new Promise((resolve) => setTimeout(resolve, 5_000));
+  }
+  await run();
```
AndrewJakubowicz, line 28: Is having this line required? It looks like this could be removed and maybe the `-1` could also be removed from the `for`. Then it will attempt to run up to 3 times.

rictic: Sometimes catching and rethrowing errors messes up error stacks, this way the error that gets thrown is untouched :)

AndrewJakubowicz (approved):
🙌 Awesome!

PR #896, opened 2023-10-01
Improve our error messages when tests fail

rictic:
This involves moving test setup into a wrapper function rigTest, which can notice failure and ask the rig to log more info when it fails.

We also modify our loggers to be able to emit to arbitrary output streams, and when testing we both use the requested logger while also emitting a detailed log to a temp file, so we can print that if the test fails.


src/logging/combination-logger.ts
```diff
@@ -0,0 +1,51 @@
...
+  getWatchLogger?(): Logger {
+    const watchLoggers = this.#loggers.map(
+      (logger) => logger.getWatchLogger?.() ?? logger,
+    );
+    return new CombinationLogger(watchLoggers, this.console);
+  }
+
+  [Symbol.dispose](): void {
+    for (const logger of this.#loggers) {
+      logger[Symbol.dispose]?.();
```
aomarks, line 47: Doesn't calling these dispose functions assume that this instance has sole ownership of the console & wrapped loggers?

src/cli-options.ts
```diff
@@ -222,6 +223,22 @@ export const getOptions = (): Result<Options> => {
     return loggerResult;
   }
 
+  let logger = loggerResult.value;
+  if (process.env['WIREIT_DEBUG_LOG_TO']) {
```
aomarks, line 227: I think I slightly prefer `WIREIT_DEBUG_LOG_FILE` over `WIREIT_DEBUG_LOG_TO`

aomarks (approved):

src/logging/combination-logger.ts
rictic: Yes. That's the intended semantics

A Console doesn't by default own its output streams though, so disposing it doesn't do anything unless the Console was constructed with the additional arg specifying that it owns its streams

src/cli-options.ts
rictic: I like that too, renamed

PR #897, opened 2023-10-01
Retry all watch tests once in case of failure.

rictic:
Watch tests seem to be fundamentally somewhat flaky as we have currently written them.

Recommend reviewing with ?W=1 because most changed lines are just whitespace changes


rictic:
I'd like our watch tests to just not be flaky, but this seems like a strict improvement

aomarks (approved):

PR #899, opened 2023-10-02
Autocomplete dependencies

rictic:


src/test/ide.test.ts
```diff
@@ -804,4 +805,410 @@ test('we can find references to a dependency', async ({rig}) => {
...
+      {
+        label: 'bar',
+        kind: completionItemKinds.dependenciesOnly,
+      },
+      {
+        label: 'files',
+        kind: completionItemKinds.filesOnly,
+      },
+      {
+        label: 'foo',
```
aomarks, line 1003: Given the suggestion `"foo"`, how does VSCode know to complete `"./child:fo|"` to `"./child:foo"` rather than to `"foo"` or `"./child:fofoo"`? I.e. how does it know to replace starting from after the `:`?

Edit: Ah, I see the answer, it must be the "triggerCharacters" setting. So it will go backwards from the cursor until it hits one of those, and then replace from that point?

aomarks (approved):
Cool!

src/test/ide.test.ts
rictic: Yeah, but it's not great at it, particularly for scripts that have a colon in their name. We have all the info to do the right thing ourselves, but I struggle to express that using the LSP. See https://github.com/microsoft/vscode/issues/194580

Something to work on down the line

PR #937, opened 2023-10-20
Fix hang when cancelling a service start up

rictic:
The promise returned by ServiceExecution.execute() would never settle if it was aborted while its dependencies were started, meaning that other dependent promises would likewise never settle.

This was made clearer by the new quiet logger, because the spinner keeps going until the top level CLI run() function returns.


rictic:
I did verify that the new test fails without the change to ServiceExecutor

aomarks (approved):

PR #942, opened 2023-10-25
Use `quiet-ci` Logger In Non-Interactive Environments

ObliviousHarmony:
We have a repository that uses PNPM and makes heavy use of the `--filter` syntax for running scripts concurrently across different projects. Each of these scripts are ran non-interactively and so the progress spinner and line writeover result in useless output. We can remedy this by checking for `process.stdout.isTTY` in addition to `process.env.CI` when deciding what the default logger should be.


README.md
```diff
@@ -854,7 +854,7 @@ The following environment variables affect the behavior of Wireit:
...
+| `WIREIT_LOGGER`         | How to present progress and results on the command line.<br><br>Options:<br><ul><li>`quiet` (default): writes a single dynamically updating line summarizing progress. Only passes along stdout and stderr from commands if there's a failure, or if the command is a service. The planned new default, please try it out.</li><li>`simple` (default): A verbose logger that presents clear information about the work that Wireit is doing.</li><li>`metrics`: Like `simple`, but also presents a summary table of results once a command is finished.</li><li>`quiet-ci` (default when env.CI or stdout.isTTY): like `quiet` but optimized for non-interactive environments, like GitHub Actions runners.</li></ul> |
```
augustjk, line 857:
```suggestion
| `WIREIT_LOGGER`         | How to present progress and results on the command line.<br><br>Options:<br><ul><li>`quiet` (default): writes a single dynamically updating line summarizing progress. Only passes along stdout and stderr from commands if there's a failure, or if the command is a service. The planned new default, please try it out.</li><li>`simple` (default): A verbose logger that presents clear information about the work that Wireit is doing.</li><li>`metrics`: Like `simple`, but also presents a summary table of results once a command is finished.</li><li>`quiet-ci` (default when env.CI or !stdout.isTTY): like `quiet` but optimized for non-interactive environments, like GitHub Actions runners.</li></ul> |
```

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
```
augustjk, line 9:
```suggestion
## Unreleased
```

README.md
ObliviousHarmony: Heh, I'm bad :smile:

rictic (approved):
This seems like an improvement.

As an aside, running multiple wireit scripts in parallel from another runner (like `pnpm --filter`) is supported but will be suboptimal, because it will invoke wireit repeatedly, run cross-package dependencies multiple times, re-check caches, schedule too much simultaneous work, etc

It will generally be more optimal to use a single runner.

ObliviousHarmony:
I know @rictic, I opened https://github.com/google/wireit/issues/944 to propose a way Wireit could be used with another runner without the performance penalty.

PR #963, opened 2023-11-03
add tests for yarn berry

appsforartists:
Currently, wireit's test suite only covers yarn classic.  This begins to include coverage for yarn berry.

This works in everything but Windows.  Two months later, and we still haven't had the time/info to make it work in Windows.  It's better to have some tests than no tests, so we're landing this for POSIX now, and tracking Windows testing in #1014.


appsforartists:
@rictic @aomarks if either of you wanna take a look at this

appsforartists:
I think this is the right direction.

We can use `yarnPath` to switch which version of `yarn` to use (which is the same that would happen in the wild).  We can even encapsulate that in something like:

```ts
const getCommandForAgent(agent) => agent.startsWith('yarn') ? 'yarn' : agent
```

and switch the current `for (const command of` to `for (const agent of`.  The only wrinkle is that we need to modify the filesystem of `rig`.  IDK if there's an immutable way to do that, or where you'd wanna put that.

PR #968, opened 2023-11-12
Add env needed by @actions/artifact

fawazahmed0:


aomarks:
This looks reasonable, but @fawazahmed0 we can't accept the PR until you have signed the Google Open Source CLA (see the failed check). Thank you!

aomarks:
Actually, on second look, I'm not sure this change alone will have the desired effect.

In Wireit, we don't use the `@actions/artifact` package. We have our own independent implementation for interfacing with the GitHub Actions caching server (https://github.com/google/wireit/blob/278613d91c771190fa9e682429804c4cd6ffe678/src/caching/github-actions-cache.ts#L54).

So, while this PR will start exposing the `ACTIONS_RUNTIME_URL` environment variable, Wireit isn't going to do anything with it, so no behavior will actually change.

@fawazahmed0 What is the behavior that you need using the `ACTIONS_RUNTIME_URL` variable? We already read the `ACTIONS_CACHE_URL` variable, and from some quick research I wasn't yet able to figure out how we might use the `ACTIONS_RUNTIME_URL` variable in Wireit.

aomarks (changes requested):
(see comment above)

aomarks:
Ah I see, so you don't use Wireit?

You could use https://github.com/crazy-max/ghaction-github-runtime if you just need the environment variables for your own purposes.

PR #1005, opened 2024-01-09
Re-Added Unhandled Exception Handling

ObliviousHarmony:
In https://github.com/google/wireit/pull/471 the handling of unhandled exceptions was removed. This adds that missing handling back so that any exceptions that aren't caught can be gracefully handled and raised to the user correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - The default logger for non-interactive environments has been switched to the 'quiet-ci' logger.
+- Unhandled exceptions will now be handled more gracefully.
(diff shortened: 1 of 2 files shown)


aomarks (changes requested):
Looks good! Mind just adding a changelog entry for this?

aomarks:
Also needs an `npm run format`

aomarks (approved):

aomarks:
> Looks good! Mind just adding a changelog entry for this?
> Also needs an npm run format

I added these, merging 👍

ObliviousHarmony:
Thanks @aomarks, I missed the changelog and am too used to the Prettier ESLint plugin :smile: Odd looking test failure though.

PR #1013, opened 2024-01-10
Release 0.14.3

aomarks:
cc @ObliviousHarmony


ObliviousHarmony (approved):
Thanks!

As an aside, technically, the last release should have been `0.15.0` since it added a new feature, right?

augustjk (approved):

augustjk:
> As an aside, technically, the last release should have been `0.15.0` since it added a new feature, right?

Wireit is still on major version 0 in which case minors would signify breaking changes and any fixes or new features that are non-breaking can be added with patches.

aomarks:
Released!

PR #1031, opened 2024-01-26
Fix bug relating to expanding symlinks to directories

aomarks:
When listing a symlink that points to a directory in `output` files, the symlink will now be directly cached as a symlink, instead of its children being cached. This also fixes an `file already exists, symlink` exception that could occur in the same situation.


augustjk (approved):
Nice! Very helpful comment.

Issue #1039, opened 2024-02-02
Upgrade Node.js in `wireit-setup-github-actions-caching` GHA


kraenhansen:
The "[wireit-setup-github-actions-caching](https://github.com/google/wireit/tree/setup-github-actions-caching/v1)" action is currently [using node16](https://github.com/google/wireit/blob/f3a3c79c553122e2fe5829eeac7d815326502903/action.yml#L5), it would be great to get this upgraded to "node20" to get rid of:

> [!WARNING]  
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: google/wireit@setup-github-actions-caching/v1. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

aomarks:
setup-github-actions-caching/v1 now uses node20

PR #1053, opened 2024-02-16
Upgrade setup-github-actions-caching to use Node 20

OrKoN:
Node 16 reached EOL last September and although there is probably okay to keep using Node 16 for this action for now, GitHub started warning about an outdated Node version being used. This PR removes that warning by upgrading to Node 20 (LTS; https://nodejs.org/en/about/previous-releases#release-schedule) and reduces the risk of breaking if Node 16 becomes not available on GitHub Actions.

Fixes https://github.com/google/wireit/issues/1039

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -2,5 +2,5 @@ name: 'wireit-setup-github-actions-caching'
 author: 'Google LLC'
 description: "Enables Wireit's GitHub Actions caching mode and exposes required environment variables"
 runs:
-  using: 'node16'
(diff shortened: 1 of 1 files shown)


OrKoN:
I have done a test run using this PR in Puppeteer https://github.com/puppeteer/puppeteer/actions/runs/7930178782/job/21651960253?pr=11934

aomarks (approved):

PR #1157, opened 2024-08-13
Fix GitHub Caching action for Windows

aomarks:
Fixes Windows support for GitHub Actions Caching, which has been broken since the security improvement in https://github.com/google/wireit/pull/1146. (This only affected users who had upgraded to `v2` of the action, and the failure was not fatal, it just disabled caching).

The problem was that under Windows, `{detached: true}` is not sufficient for keeping the server alive after the parent shell exits. The parent shell will exit as soon as the main launching script is done, because the action is invoked through a GitHub Actions `uses:` clause (as opposed to `run:` where we would be sharing our parent shell with subsequent steps).

We now use `start` so that the server is not killed when our parent shell exits: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/start.

Also listens only on `localhost` instead of all hosts (shouldn't matter because we should have strong network isolation for free through the GitHub Actions container, but seems wise, and good in case this code runs in some other context for some reason).

Also adds a `CHANGELOG.md` file.

### DIFF
--- a/main.js
+++ b/main.js
@@ -4,16 +4,84 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { fork } from "node:child_process";
+import { spawn } from "node:child_process";
+import { mkdtempSync, openSync, readFileSync } from "node:fs";
+import { tmpdir } from "node:os";
 import { join } from "node:path";
 
 console.log("[main] Launching background custodian service");
-const server = fork(join(import.meta.dirname, "custodian.js"), {
-  detached: true,
-  stdio: "inherit",
-});
 
-server.on("message", (status) => {
-  console.log("[main] Received status from custodian service", status);
-  process.exit(status);
-});
+const scriptPath = join(import.meta.dirname, "custodian.js");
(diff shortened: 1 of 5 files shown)


main.js
```diff
@@ -4,16 +4,84 @@
...
+  await new Promise((resolve) => setTimeout(resolve, pollMillis));
+  stdout = readFileSync(outPath, "utf8");
+  stderr = readFileSync(errPath, "utf8");
+  if (stdout.match(/Listening on port/) !== null) {
+    console.error(stderr);
+    console.log(stdout);
+    console.log(`[main] Custodian server ready`);
+    process.exit(0);
+  }
+  if (Date.now() - start > timeoutSecs * 1000) {
```
rictic, line 79: Mild optional thing for the future, it's often a bit more elegant to do:

```ts
const timedOutTime = Date.now() + (timeoutSecs * 1000);
```

rictic (approved):

main.js
aomarks: Ah yes, nice. Done.

PR #1176, opened 2024-08-31
Add support for filesystem polling

aomarks:
Adds `WIREIT_WATCH_STRATEGY` and `WIREIT_WATCH_POLL_MS` environment variables which can be used to switch filesystem detection from event-based to polling:

```
export WIREIT_WATCH_STRATEGY=poll
export WIREIT_WATCH_POLL_MS=<a number if you don't like the default of 500>
```

Currently this just sets the chokidar `usePolling` and `interval` options. It might be nice to actually bypass chokidar entirely so that this option can be used to bypass any potential issues with chokidar generally, but it's slightly complex (just re-running the script on an interval isn't great, because of scripts which aren't set up well for caching).

This is published as `v0.14.9-pre.1`. Would be great to hear from @peschee if this is a useful workaround for the OOM errors they have been seeing in their project.


peschee:
So cool @aomarks, will give this a spin on Monday.

CHANGELOG.md
```diff
@@ -6,7 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Added
+
+- Add support for forcing the use of filesystem polling instead of OS events in watch mode. Set the environment variable `WIREIT_WATCH_STRATEGY=poll`, and optionally the `WIREIT_WATCH_POLL_MS` (default 500).
```
peschee, line 13:
```suggestion
- Add support for forcing the use of filesystem polling instead of OS events in watch mode. Set the environment variable `WIREIT_WATCH_STRATEGY=poll`, and optionally the `WIREIT_WATCH_POLL_MS` (default `500`).
```

peschee:
@aomarks I gave this a go yesterday and in my first tests, using `WIREIT_WATCH_STRATEGY=poll` I'm not experiencing any OOM issues anymore. I will keep an eye out the next few days, but from where I stand, this looks like a great addition to me!

src/cli-options.ts
```diff
@@ -448,3 +451,47 @@ function parseRemainingArgs(
...
+
+const DEFAULT_WATCH_STRATEGY = {strategy: 'event'} as const;
+const DEFAULT_WATCH_INTERVAL = 500;
+
+/**
+ * Interpret the WIREIT_WATCH_* environment variables.
+ */
+function readWatchConfigFromEnv(): Options['watch'] {
+  switch (process.env['WIREIT_WATCH_STRATEGY']) {
+    case 'event':
```
justinfagnani, line 463: Style nit: I definitely prefer avoid fall-through myself. Would this be easier as an if/else statement?

src/cli-options.ts
```diff
@@ -448,3 +451,47 @@ function parseRemainingArgs(
...
+    case undefined: {
+      return DEFAULT_WATCH_STRATEGY;
+    }
+    case 'poll': {
+      let interval = DEFAULT_WATCH_INTERVAL;
+      const intervalStr = process.env['WIREIT_WATCH_POLL_MS'];
+      if (intervalStr) {
+        const parsed = Number(intervalStr);
+        if (Number.isNaN(parsed) || parsed <= 0) {
+          console.error(
```
justinfagnani, line 474: If this is continuing to operate with the fallback value, should this be `console.warn()` instead?

src/cli-options.ts
```diff
@@ -448,3 +451,47 @@ function parseRemainingArgs(
...
+          interval = parsed;
+        }
+      }
+      return {
+        strategy: 'poll',
+        interval,
+      };
+    }
+    default: {
+      console.error(
```
justinfagnani, line 489: Same here.

justinfagnani (approved):

src/cli-options.ts
aomarks: This whole file (and other files) uses this style of switching on `state`. We're doing exhaustive checking for every state for every transition. I actually kinda like using fallthrough here, it's just the most concise way to write out all the permutations.

src/cli-options.ts
aomarks: Both go to stderr, but in some configurations warnings might get filtered out? I guess because it's definitely something wrong, I think it should be an error to maximize the chance of somebody seeing it.

peschee:
💪🏼

PR #1185, opened 2024-09-07
Update deps; New eslint config; Fix budgeted fs wrapper bug

aomarks:
Update all the dependencies to latest major versions. Only eslint required any changes.

Eslint 9 has a new config file format, so I switched to that. I also adopted the new recommended style for typescript-eslint which has some new rules. The new rules found a few style issues, but also a real issue.

The real issue was that an `await` was being omitted in the budgeted fs module. We could also update to `using`, but let's do that in another PR for the whole project.


eslint.config.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+import pluginJs from '@eslint/js';
+import noOnlyTests from 'eslint-plugin-no-only-tests';
+import tseslint from 'typescript-eslint';
+
+export default tseslint.config(
+  {
+    ignores: [
+      '**/*.js',
```
justinfagnani, line 14: I'm surprised that tseslint isn't already ignoring files not in the tsconfig. Is there a way to do that?

Otherwise, might this help? https://github.com/antfu/eslint-config-flat-gitignore

eslint.config.js
```diff
@@ -0,0 +1,59 @@
...
+      '**/.wireit/',
+      '**/node_modules/',
+      'lib/',
+      'vscode-extension/.vscode-test/',
+      'vscode-extension/lib/',
+      'vscode-extension/built/',
+    ],
+  },
+  {
+    files: ['**/*.ts'],
```
justinfagnani, line 27: should this be `src/**/*.ts`?

eslint.config.js
```diff
@@ -0,0 +1,59 @@
...
+    languageOptions: {
+      parser: tseslint.parser,
```
justinfagnani, line 29: I didn't think you had to set the parser manually anymore.

Here's a recent config I made when upgrading to ESLint v9:

```ts
import tseslint from 'typescript-eslint';
import eslint from '@eslint/js';
import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended';

export default [
  eslint.configs.recommended,
  eslintPluginPrettierRecommended,
  ...tseslint.configs.recommendedTypeChecked,
  {
    languageOptions: {
      parserOptions: {
        projectService: true,
        tsconfigRootDir: import.meta.dirname,
      },
    },
    rules: {
      "@typescript-eslint/no-non-null-assertion": "error"
    },
  },
];
```

It doesn't use `tseslint.config()` because the config is in .js and doesn't need a type-safe config merger.

eslint.config.js
aomarks: Do you use a path argument when you run eslint in your npm script, like `eslint src/**/*.ts` or whatever?

I like to set eslint up so that you can run it with no path argument. My experience is that when you run eslint with no path argument, it will try and lint all supported files. The tseslint config may only apply to typescript project files, but eslint is still going to try and lint everything else with the non-tseslint rules.

So I think it's still good to have all your non-lintable files listed here.

Syncing with .gitignore makes sense, but I think I'm going skip that plugin since the structure changes so rarely and I don't want to add a dep. I did clean up the list, though.

eslint.config.js
aomarks: I think `**/*.ts` is actually fine, because if we're globally ignoring all non-source files, then a general ".ts extension" filter seems perfectly fine.

This way we don't have to know anything about the project layout, which in this case would also require that we add `vscode-extension/src`.

eslint.config.js
aomarks: > I didn't think you had to set the parser manually anymore.

Huh, you're right! I think I was getting mislead by a combination of things:

- My config was initially matching .js files not covered by a TypeScript config because I didn't have the right `files` filter.
 
- eslint then tried to run a TypeScript-parser-depending rule on that js file, but that rule has an outdated error message saying that you need to set a parser. But the real problem was just that it shouldn't have been covered by the TypeScript rule at all.


> It doesn't use tseslint.config() because the config is in .js and doesn't need a type-safe config merger.

Ah nice. Didn't realize this was just a typings thing.

> Here's a recent config I made when upgrading to ESLint v9:

Cool! I adopted some of this.

The main difference now is that I have this `{files: ["**/*.ts"]}` constraint on all the TypeScript setting objects (including the tseslint ones). What I found is that I otherwise get errors with `projectService: true` where I get errors about tseslint not being able to find a `tsconfig.json` for the one or two `.js` files we have here. A little odd, but I can't find a simpler arrangement after fiddling for a while -- I always end up with some kind of error relating to non-ts  files.

aomarks: >   eslintPluginPrettierRecommended,

Oh also, curious why you like using this prettier eslint plugin? (instead of just using the prettier binary).

PR #1271, opened 2025-01-28
Release 0.14.10

aomarks:
Also update dependencies.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -18,7 +18,8 @@
     "noImplicitOverride": true,
     "incremental": true,
     "tsBuildInfoFile": ".tsbuildinfo",
-    "composite": true
+    "composite": true,
+    "skipLibCheck": true
   },
   "include": ["src/**/*.ts"],
   "exclude": []
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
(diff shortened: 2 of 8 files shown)


tsconfig.json
```diff
@@ -18,7 +18,8 @@
     "noImplicitOverride": true,
     "incremental": true,
     "tsBuildInfoFile": ".tsbuildinfo",
-    "composite": true
+    "composite": true,
+    "skipLibCheck": true
```
rictic, line 22: why did this become necessary?

rictic (approved):

tsconfig.json
aomarks: These and a bunch of similar tsc errors from `vscode-jsonrpc`

```
node_modules/vscode-jsonrpc/lib/common/linkedMap.d.ts:28:5 - error TS2416: Property 'forEach' in type 'LinkedMap<K, V>' is not assignable to the same property in base type 'Map<K, V>'.
  Type '(callbackfn: (value: V, key: K, map: LinkedMap<K, V>) => void, thisArg?: any) => void' is not assignable to type '(callbackfn: (value: V, key: K, map: Map<K, V>) => void, thisArg?: any) => void'.
    Types of parameters 'callbackfn' and 'callbackfn' are incompatible.
      Types of parameters 'map' and 'map' are incompatible.
        Type 'LinkedMap<K, V>' is not assignable to type 'Map<K, V>'.
          The types returned by 'entries()' are incompatible between these types.
            Property '[Symbol.dispose]' is missing in type 'IterableIterator<[K, V]>' but required in type 'MapIterator<[K, V]>'.

28     forEach(callbackfn: (value: V, key: K, map: LinkedMap<K, V>) => void, thisArg?: any): void;
       ~~~~~~~
```

PR #1280, opened 2025-02-07
fix: provide a prefix to GitHub Action cache

Lightning00Blade:
Hey I was having some issue with GitHub actions, one thing I wanted to eliminate for the cause is the WireIt cache.
Problem is there is no way to filter out caches besides the Cache id. This means that I needed to nuke all the caches if you want to remove it, which is sub-optimal as if you are fixing an issue you want to do it 1 step at a time.

_Send through github.dev so couldn't run tests_


aomarks (approved):
Makes sense. Thanks for the PR!

aomarks:
Released as 0.14.11

PR #1288, opened 2025-02-20
New dependency parser and introduce `#` delimiter

aomarks:
This PR introduces `#` as the new and preferred way to delimit the package and script portions of a dependency specifier, by way of a new dependency specifier parser. The `:` delimiter will remain supported indefinitely.

The goal of the new parser, and the new `#` delimiter, is to prepare for adding a number of new features to dependency specifiers, such as https://github.com/google/wireit/issues/23 (but see `dependency-parser.ts` for the actual list, the syntax has changed, I will update the issue). None of the new features are implemented in this PR, they will come in followups.

`#` was chosen as the new preferred delimiter because `:` is extremely common in npm script names, whereas `#` is uncommon in both paths and npm script names. This will allow us to differentiate between `npmpackage:script` (which becomes `npmpackage#script`) and `localscript:withcolon`. `#` is also the delimiter used by Turborepo, which seems like a slightly nice alignment.

### DIFF
--- a/src/test/dependency-parser.test.ts
+++ b/src/test/dependency-parser.test.ts
@@ -1,0 +1,350 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {suite} from 'uvu';
+import * as assert from 'uvu/assert';
+import {
+  parseDependency,
+  type ParsedPackage,
+  type ParsedScript,
+} from '../analysis/dependency-parser.js';
+import type {DiagnosticWithoutFile} from '../error.js';
+
+const test = suite<object>();
+
+const cases: Array<
+  [
+    /**
+     * Dependency specifier, e.g. "./pkg#scr".
+     */
+    string,
+
+    /**
+     * A string the same length as the dependency specifier, where each
+     * character represents how we should interpret the corresponding character
+     * from the dependency, with P = package, S = script, E = error, and
+     * anything else is ignored (e.g. "PPPPP_SSS"). This way we can clearly
+     * annotate the expected character ranges for each parsed component.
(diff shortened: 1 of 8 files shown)


src/test/dependency-parser.test.ts
```diff
@@ -0,0 +1,358 @@
...
+      package: {
+        kind: 'path',
+        path: './pkg{foo,bar}',
+      },
+      script: {
+        kind: 'name',
+        name: 'scr{foo,bar}',
+      },
+    },
+  ],
```
augustjk, line 273: if/when `*` and `{,}` expands are supported, they wouldn't actually require backslash escapes, would they? these tests are simply demonstrating handling of unnecessary escapes?

src/test/dependency-parser.test.ts
```diff
@@ -0,0 +1,358 @@
...
+    String.raw`./pkg#\#scr\#`,
+    String.raw`PPPPP_SSSSSSS`,
+    {
+      package: {kind: 'path', path: './pkg'},
+      script: {kind: 'name', name: '#scr#'},
+    },
+  ],
+  [
+    String.raw`\./scr`,
+    String.raw`_SSSSS`,
```
augustjk, line 120: shouldn't the leading `\` for escaping considered part of the range?

src/test/dependency-parser.test.ts
```diff
@@ -0,0 +1,358 @@
...
+    {
+      package: {kind: 'this'},
+      script: {kind: 'name', name: './scr'},
+    },
+  ],
```
augustjk, line 125: perhaps an extra test case where leading escaped period `\.` with a `#` or `:` delimiter makes the first segment be interpreted as an npm package?

Issue #1297, opened 2025-03-24
Time sensitive: GitHub Actions cache service shutdown


Link-:
The actions/cache backend service has been rewritten from the ground up for improved performance and reliability. The new service has been gradually rolling out since February 1st, 2025. The legacy service **will be sunset by April 15th 2025**.

We have identified that this project is integrating with the legacy cache service without using the official and supported [package](https://github.com/actions/toolkit/tree/main/packages/cache). Unfortunately, that means that you have to introduce code changes in order to be compatible with the new service we're rolling out.

The new service uses an entirely new set of internal API endpoints. To help with your changes we have provided the proto definitions below to help generate compatible clients to speed up your migration. 

**These internal APIs were never intended for consumption the way your project is at the moment. Since this is not a paved path we endorse, it's possible there will be breaking changes in the future. We are reaching out as a courtesy because we do not wish to break the workflows dependent on this project.**

Please introduce the necessary changes ASAP. Otherwise, storing and retrieving cache entries will start to fail. There will be no need to offer backward compatibility as the new service has been rolled out to all repositories.

The cutover point in time will be the moment the new service is rolled out to a given repository. Then, users will no longer have access to cache entries from the legacy service and are expected to store / retrieve cache entries from the new service.

<details>
  <summary>Expand for proto definitions</summary>

#### cache.proto
```
syntax = "proto3";

import "cachemetadata.proto";

package v1;

service CacheService {
    // Generates a SAS URL with write permissions to upload a cache archive
    rpc CreateCacheEntry(CreateCacheEntryRequest) returns (CreateCacheEntryResponse);
    // Indicate the completion of a cache archive upload. Triggers post-upload processing
    rpc FinalizeCacheEntryUpload(FinalizeCacheEntryUploadRequest) returns (FinalizeCacheEntryUploadResponse);
    // Generates a SAS URL with read permissions to download a cache archive
    rpc GetCacheEntryDownloadURL(GetCacheEntryDownloadURLRequest) returns (GetCacheEntryDownloadURLResponse);
}

message CreateCacheEntryRequest {
    // Scope and other metadata for the cache entry
    results.entities.v1.CacheMetadata metadata = 1;
    // An explicit key for a cache entry 
    string key = 2;
    // Hash of the compression tool, runner OS and paths cached
    string version = 3;
}

message CreateCacheEntryResponse {
    bool ok = 1;
    // SAS URL to upload the cache archive
    string signed_upload_url = 2;
}

[...]

PR #1302, opened 2025-04-10
Support V2 GitHub Caching service (Part 2/2: Wireit)

aomarks:
Add support for the new V2 version of the GitHub caching API.

Requires https://github.com/google/wireit/pull/1301 to land first, which provides a new environment variable we need to read.

Also adds tests against the real production service in CI.

See https://github.com/google/wireit/issues/1297 and https://github.blog/changelog/2025-03-20-notification-of-upcoming-breaking-changes-in-github-actions/#decommissioned-cache-service-brownouts for more background on the V2 changes.


src/caching/github-actions-cache.ts
```diff
@@ -193,27 +206,41 @@ export class GitHubActionsCache implements Cache {
...
     const version = this.#computeVersion(fingerprint);
     const key = this.#computeCacheKey(script);
-    const url = new URL('_apis/artifactcache/cache', this.#baseUrl);
-    url.searchParams.set('keys', key);
-    url.searchParams.set('version', version);
-
-    using requestResult = this.#request(url);
+    const url = new URL(
+      // See https://github.com/actions/toolkit/blob/930c89072712a3aac52d74b23338f00bb0cfcb24/packages/cache/src/generated/results/api/v1/cache.twirp-client.ts#L91
+      'twirp/github.actions.results.api.v1.CacheService/GetCacheEntryDownloadURL',
```
rictic, line 211: Could this be an absolute path?

rictic (approved):

src/caching/github-actions-cache.ts
aomarks: In the V1 API, the base URL we were passed came with a path component, so it was important not to clobber that by constructing an API URL with an absolute path. The V2 base URL is just `https://results-receiver.actions.githubusercontent.com/`, but I think if it were to change to come with a path, I would want to preserve it.

PR #1310, opened 2025-04-30
Remove support for Node 18 which is now EOL

aomarks:
It's Node 18 EOL day! Also updates deps and CI image versions.


CHANGELOG.md
```diff
@@ -8,6 +8,17 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Changed
+
+- **[BREAKING]** Node 18 is no longer supported because it has reached
+  end-of-life. Node 20 is now the oldest supported version. See
+  https://github.com/nodejs/Release#release-schedule.
+
+  Note we will no longer issue breaking semver increments for changes to
```
rictic, line 17: Not sure how I feel about that. Why?

aomarks: > Not sure how I feel about that. Why?

Benefits to incrementing:

- Users on unsupported versions of Node can continue using Wireit without interruption (which would be updating their Node or pinning their Wireit version).

Costs to incrementing:

- Users are less likely to receive our bug fixes and security patches, because updating to a new major version is usually a manual action that users are slow to perform.

---

I also offer this principle (which I hold lightly):

If you are maintaining a Node project, you have the baseline responsibility to use a supported version of Node, just as you have a responsibility to use a supported version of your operating system. It should not be surprising that if you are using an unsupported version of Node, the dependencies you download from NPM might not work properly. If you're running an unsupported Node, you have the bigger problem of not getting security updates anymore, and it should not the responsibility of NPM package maintainers to help you defer doing that important update.

---

Thinking about alternative strategies...

Things we could do when a Node version becomes unsupported:

1. Update the documented Node versions we support in the README.
2. Increment the `engines` field in the `package.json`. (This warns, but doesn't block, users who are installing on older Node versions.)
3. Stop testing the unsupported version in CI.
4. Start using Node features not available in the unsupported version (this is of course what *actually* breaks users).
5. Major increment our version.

Some strategies we could adopt:

A. 1-4 (this PR)
B. 1-5
C. 1+2, wait for a grace period (e.g. 3 months), then 3+4
D. Don't actually do anything (other than maybe 1) unless there is actually a new Node feature we want to start using, and in that case do 1-5.

Issue #1346, opened 2025-12-21
Watch mode will not be settable from command-line after npm 12 flag parsing changes


aomarks:
According to https://github.com/npm/cli/pull/8071 (which adds a warning), npm will change the way arguments are parsed in a way that will break the way wireit looks for the `--watch` environment variable.

Previously, `npm run build --foo` would set the `npm_config_foo` environment variable, and we relied upon that fact for detecting when the `--watch` flag had been sent.

In npm 12, this will be an error about "watch" being an unrecognized npm argument. This seems like a good change for npm (silently eating arguments does seem bad), but I don't believe we currently have another way to enable wireit watch mode.

Proposals:

1. Add a `WIREIT_WATCH=1` environment variable.
2. Special case `--wireit-watch` (and perhaps reserve `--wireit-` prefixed flag) to be extracted from the additional arguments that are usually passed down to the actual command (e.g. `npm run build -- --wireit-watch --flag-for-build`)
3. Solve for https://github.com/google/wireit/issues/553 to allow a script to be configured as intrinsically watching (e.g. a `{"watch":true}` config setting), so that people can easily define e.g. `build:watch` scripts instead of using flags in the first place.

I'm leaning towards doing 1 in the short term, because we already handle all other runtime wireit settings with `WIREIT_` environment variables and it's very easy, and 3 as a followup.

PR #1418, opened 2026-05-19
feat(cli): accept WIREIT_WATCH env var as a --watch escape hatch (#1346)

kiwigitops:
Resolves the short-term proposal in #1346.

## Problem

[npm/cli#8071](https://github.com/npm/cli/pull/8071) lands a warning today and will turn into a hard error in npm 12: `npm run build --watch` is rejected because `--watch` is not a recognized npm argument. Until now, wireit relied on npm's old behaviour of silently turning `npm run build --watch` into `npm_config_watch=` in the script's environment. After npm 12 there is no way to enable wireit watch mode from `npm run` — neither `--watch` (rejected) nor `-- --watch` (forwarded to the underlying script, not wireit).

@aomarks proposed in #1346:

> 1. Add a `WIREIT_WATCH=1` environment variable.
> [...]
> I'm leaning towards doing 1 in the short term, because we already handle all other runtime wireit settings with `WIREIT_` environment variables and it's very easy [...]

This PR implements proposal 1.

## Change

Two `getArgvOptions` call sites now recognize `WIREIT_WATCH` (any non-undefined value) as an additional trigger for watch mode, OR'd with the existing detection:

```diff
   case 'npm': {
     return {
       watch:
-        process.env['npm_config_watch'] !== undefined
+        process.env['npm_config_watch'] !== undefined ||
+        process.env['WIREIT_WATCH'] !== undefined
           ? readWatchConfigFromEnv()
           : false,
       extraArgs: process.argv.slice(2),
     };
   }
```

And the `parseRemainingArgs` helper (used for yarn / yarnBerry / pnpm / nodeRun) gets a matching post-loop check so the escape hatch works on every agent, not just npm:

```diff
+  // Also accept WIREIT_WATCH as an escape hatch (matches the npm branch behaviour,
+  // useful for yarn/pnpm/node --run wrappers that scrub --watch from argv).
+  if (!watch && process.env['WIREIT_WATCH'] !== undefined) {
+    watch = readWatchConfigFromEnv();
+  }
   return {
     watch,
     extraArgs,
   };
```

The existing `WIREIT_WATCH_STRATEGY` / `WIREIT_WATCH_POLL_MS` env vars still control the watch strategy via the unchanged `readWatchConfigFromEnv()`.

## Test plan

- [x] Added an `${agent} WIREIT_WATCH=1 run` test case alongside the existing `WIREIT_LOGGER` env-var tests. Runs in the per-agent loop, so it covers npm, yarnClassic, yarnBerry, pnpm, and `node --run` in one go.
- [x] All existing tests that exercise `--watch` and `WIREIT_WATCH_STRATEGY` are untouched and should continue to pass — the change is purely additive (an extra OR'd condition).

## Notes for reviewers

- I deliberately scoped this to proposal 1 only. Proposals 2 (`--wireit-watch` arg) and 3 (declarative `{"watch": true}`) are bigger surface-area changes and can land on their own.
- `WIREIT_WATCH=` (empty string) also triggers watch mode, because the existing detection uses `!== undefined` rather than a truthy check — kept consistent so behaviour is predictable; matches how `WIREIT_DEBUG_LOGGER` is detected in `logger.ts` (via `Boolean(...)`) is intentionally NOT used here so an empty `WIREIT_WATCH=` set in CI's env still triggers, mirroring `npm_config_watch=`.


src/cli-options.ts
```diff
@@ -276,9 +276,15 @@ function getArgvOptions(
...
+      //
+      // npm 12 (https://github.com/npm/cli/pull/8071)
+      //   - "npm run build --watch" is rejected as an unrecognized npm argument
+      //     instead of being silently turned into "npm_config_watch", so we also
+      //     accept WIREIT_WATCH=1 as an escape hatch (see #1346).
       return {
         watch:
-          process.env['npm_config_watch'] !== undefined
+          process.env['npm_config_watch'] !== undefined ||
+          process.env['WIREIT_WATCH'] !== undefined
```
aomarks: Nit, but it seems like we don't need the check on this line because it's going to get set regardless later on for all runners.

src/cli-options.ts
```diff
@@ -446,6 +452,11 @@ function parseRemainingArgs(
         `e.g. "npm run build -- --extra".`,
     );
   }
+  // Also accept WIREIT_WATCH as an escape hatch (matches the npm branch behaviour,
+  // useful for yarn/pnpm/node --run wrappers that scrub --watch from argv).
+  if (!watch && process.env['WIREIT_WATCH'] !== undefined) {
```
aomarks: Let's do `process.env['WIREIT_WATCH']?.toLowerCase() === 'true'` (because as written, if you wrote `WIREIT_WATCH=false` it would currently get set to true)

aomarks (changes requested):
Thank you for the PR! If you don't mind, could you

- Rebase against main which brought in some test updates recently?
- Update the `CHANGELOG.md` file to mention the new variable?

src/cli-options.ts
aomarks: Oh, just noticed that this function doesn't get called in the npm case. That's fine, but could I ask that we add a test case for all the runners, since it should apply in all cases? (And if you can find a natural spot to check the variable in just one place instead of 2, that would be nice)

aomarks (changes requested):
Sorry, one more request: could you update the README section to mention the variable in the "Watch mode" section, and also in the "Environment variables" reference section?

aomarks (approved):
Thanks!

PR #1433, opened 2026-06-11
Replace uvu with node:test (24/24)

bennypowers:
## Summary

Completes the test framework migration started in #1369 by converting
the remaining 12 test files and 2 shared utilities from uvu to
node:test, then removing the uvu dependency entirely.

- Convert all remaining test files to use `node:test` and `node:assert`
- Map uvu-specific assertions to node:assert equivalents, preserving
  exact semantics (`assert.not` to `assert.ok(!(...))`; `assert.equal`
  on objects/arrays to `assert.deepEqual`; `assert.is` to
  `assert.strictEqual`; etc.)
- Consolidate `rigTestNode` into `rigTest` with optional `ms` and `env`
  options for timeout and environment variable forwarding
- Replace the vscode extension's uvu-based test entrypoint with a
  minimal runner that imports and executes exported test functions
- Fix timer leak in `rigTest` timeout (clearTimeout on settle)
- Remove dead `NODE_MAJOR_VERSION < 16` branch in `check-script-output`
- Update all wireit commands from `uvu` to `node --test`
- Remove uvu from devDependencies

One service test ("caching with service dependencies works in watch
mode") hangs under node:test's runner and is skipped pending
investigation.

## Test plan

- [ ] `npm run lint` passes
- [ ] `npm run format:check` passes
- [ ] `npm test` passes (all 31 scripts, 30 ran + 1 cached)
- [ ] Verify no `uvu` imports remain: `grep -r "from 'uvu" --include='*.ts' src/ vscode-extension/src/`
- [ ] Investigate skipped service test in follow-up


aomarks:
Looks like Node 18 doesn't like `--test-force-exit`

aomarks (approved):
Looks good, thank you @bennypowers !

PR #1439, opened 2026-06-23
Update dependencies and documentation and release 0.14.13

aomarks:


src/caching/github-actions-cache.ts
```diff
@@ -641,7 +641,9 @@ ${blockIds.map((blockId) => `  <Uncommitted>${blockId}</Uncommitted>`).join('\n'
         ],
         (error: unknown) => {
           if (error != null) {
-            reject(new Error(`tar error: ${String(error as Error)}`));
+            const message =
+              error instanceof Error ? error.message : JSON.stringify(error);
+            reject(new Error(`tar error: ${message}`));
```
rictic, line 646: Here and elsewhere, consider:

```ts
new Error(`tar error: ${message}`, {cause: error})
```

src/language-server.ts
```diff
@@ -28,7 +33,7 @@ import {IdeAnalyzer} from './ide.js';
 const ideAnalyzer = new IdeAnalyzer();
 const connection = createConnection(ProposedFeatures.all);
 
-connection.onInitialize((init) => {
+connection.onInitialize((init: InitializeParams) => {
```
rictic: Huh, surprising! Why do we need to annotate these all of a sudden?

rictic (approved):

src/caching/github-actions-cache.ts
aomarks: Done

src/language-server.ts
aomarks: Oops, we didn't actually need them, it was a red herring from a different error migrating to ts 6.

PR #1444, opened 2026-06-25
Upgrade to chokidar 4 and restore glob support

aomarks:
Chokidar 4 dropped support for globs, so we had to re-implement that.

Chokidar 5 is also out, but it drops support for Node 18, so I'm going to take that bump for a breaking release after this one.

We already had our own test suite for globbing, I added a couple more relating to !negations, and made sure it passed before and after the upgrade.

### DIFF
--- a/src/util/chokidar-with-globs.ts
+++ b/src/util/chokidar-with-globs.ts
@@ -1,0 +1,155 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import chokidar, {
+  type ChokidarOptions,
+  type FSWatcher as ChokidarFSWatcher,
+} from 'chokidar';
+import picomatch from 'picomatch';
+import globParent from 'glob-parent';
+import * as pathlib from 'path';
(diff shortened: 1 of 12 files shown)


src/util/chokidar-with-globs.ts
```diff
@@ -0,0 +1,155 @@
...
+      },
+    },
+  );
+
+  // Chokidar 4 only checks `ignored` during the initial directory scan, not
+  // on subsequent change events. Override `emit` to also filter those.
+  const originalEmit = watcher.emit.bind(watcher);
+  watcher.emit = ((event: string, ...args: unknown[]): boolean => {
+    // For individual events like 'add' or 'change', the path is args[0].
+    // For the 'all' meta-event, the path is args[1] (args[0] is the event name).
```
rictic, line 129: How does this handle a file named `all`?

rictic (approved):

src/util/chokidar-with-globs.ts
aomarks: It works ok because path is positional. Added tests to verify.

PR #1480, opened 2026-08-24
Use http or https for GitHub Actions cache based on URL scheme

### LINKED ISSUE #1479: GitHub Actions cache client rejects http:// ACTIONS_RESULTS_URL
`request()` in `src/caching/github-actions-cache.ts` always calls `https.request()`. If `ACTIONS_RESULTS_URL` is `http://`, Node throws before any connection is made:

```
TypeError [ERR_INVALID_PROTOCOL]: Protocol "http:" not supported. Expected "https:"
```

GitHub-hosted runners serve the cache API over HTTPS, so this does not show up there. Some third-party runners rewrite `ACTIONS_RESULTS_URL` to an `http://` cache proxy (we hit this on Depot). `@actions/http-client` already picks the transport from `parsedUrl.protocol === 'https:'`.

I have a small fix ready: choose `http` vs `https` from the URL scheme, and a fake-server test that speaks HTTP.

maciejb:
Fixes #1479

## Problem

`request()` always calls `https.request()`. Node throws `ERR_INVALID_PROTOCOL` when `ACTIONS_RESULTS_URL` is `http://`, before any socket is opened.

GitHub-hosted runners use HTTPS, so this never comes up there. Some third-party runners rewrite `ACTIONS_RESULTS_URL` to an `http://` cache proxy. We hit this on Depot: wireit failed on the first cache lookup.

`@actions/http-client` already chooses the transport with `parsedUrl.protocol === 'https:'`.

## Change

Parse the request URL and call `http.request` or `https.request` from the scheme. Nothing else about the cache API changes.

The fake GitHub cache server now skips TLS when no cert is passed, so a test can speak HTTP. Existing tests still pass a cert and stay on HTTPS.

## Test plan

- [x] New fake-server test: HTTP `ACTIONS_RESULTS_URL`, two runs, second run is a cache hit (`getBlob: 1`, command not invoked again).
- [x] `npm run test:cache-github-fake` (26/26, including the existing HTTPS cases)


aomarks (approved):
Thanks for the PR!

PR #1481, opened 2026-08-26
fix: lru local cache eviction #71

### LINKED ISSUE #71: Garbage collection for cache directory
The `.wireit/<script>/cache` directory currently can grow indefinitely. We should implement a garbage collection strategy to cap the size of this directory.

An LRU cache with a configurable maximum number of entries seems like what we want.

We will want to make sure we have an efficient way to maintain the cache hit rate data which scales well with the size of the cache. We will probably want some kind of on-disk index file that lets us read/write cache hit rates efficiently, to determine which cache entry needs to be deleted when the cap is hit. A doubly-linked-list implemented in the filesystem itself with symlinks (or just files containing SHAs) could also be an interesting way to do this.

asmyshlyaev177:
# Limit local cache size with per-script LRU eviction

Fixes #71

## Problem

`.wireit/<script>/cache/<fingerprint>/` gains a full copy of the script's
`output` per fingerprint, and nothing removes them. The README told users to
`rm -rf .wireit/*/cache` by hand.

Cheap when `output` is `dist/`, expensive when it isn't. A repo declaring
`node_modules` as output reached 73 GB at ~540 MB per entry.

## Change

`LocalCache` keeps the N most recently *used* entries per script — written by
`set()`, read by `get()`, or relied upon by a fresh script. Eviction runs after
each successful `set()`.

- Default N = **10**. `WIREIT_CACHE_MAX_ENTRIES` overrides: positive integer, or
  `infinity` for the old behavior. Parsed like the existing `WIREIT_PARALLEL`.

### Recency = directory mtime

No index file, no consistency problem. `atime` is unusable: filesystems are
commonly mounted `noatime` or `relatime`.

Three things stamp an entry, and the third is the one that isn't obvious:

- `set()` writes it.
- `get()` stamps the entry it returns, so a frequently restored entry isn't
  evicted for being old — that's LRU rather than FIFO.
- `markEntryRecentlyUsed()`, a new `Cache` method, stamps the entry a *fresh*
  script relied upon. Nothing is restored on that path, so `get()` is never
  called (see the fresh check in `standard.ts`), and without this an
  always-fresh script's entry would look untouched and eventually be evicted.
  It is a no-op for `GitHubActionsCache`, which doesn't choose its own evictions.

The entry `set()` just wrote is excluded from the eviction candidates rather
than relying on its mtime to save it — mtime resolution is coarse on some
filesystems, so it can tie with an older entry and lose the sort.

### Eviction is a rename; the deleting happens at the end of the run

An entry is a full copy of `output`, so deleting one can take a while. A script
shouldn't wait for that.

`set()` blocks only on `rename`-ing each doomed entry into the package's
`.wireit/trash/<16 random hex>`. A new `sweepTrash(signal?)` on the `Cache`
interface empties that folder:

- `cli.ts` awaits it once the run is over (and once the watcher exits), and
  prints "Ctrl-C is safe" if it is still going after a second. `SIGINT` and
  `SIGTERM` abort it, and anything left is picked up by the next run.
- Watch mode also sweeps in the background after each iteration, so a long
  session doesn't accumulate trash.
- Deletes tolerate `ENOENT`: two Wireit processes can be sweeping the same
  folder at once, which is also why sweeping takes no lock.

The trash name is random rather than derived from the entry, for two reasons.
The same entry can be evicted, written and evicted again before a sweep reaches
it. And every file in the entry is renamed onto that path, so it needs to be
short: `.wireit/trash/<16>` is 59 characters shorter than the
`.wireit/<script>/cache/<64>` it came from, so a tree that fit before still
fits, including against the Windows path limit.

### Locking

None added. Eviction touches only the calling script's own cache folder, and
`StandardScriptExecution#acquireSystemLockIfNeeded` already holds that script's
lock across `get()` and `set()`. It skips the lock only for an empty `output`,
where entries are empty directories.

Sweeping is deliberately unlocked, per the point about `ENOENT` above.

### Housekeeping never fails a script

Stamping, eviction and sweeping swallow errors:

- an unstampable directory (read-only mount, foreign owner) must still return a
  hit
- an unmovable or undeletable entry (EPERM/EBUSY on Windows) must not fail a
  script that is already cached — and the renames use `Promise.allSettled`, so
  one stuck entry doesn't block evicting the rest

Degraded behavior is "cache larger than requested", not a failed build.

Entries are ranked with `lstat`, not `stat`, so a broken symlink in the folder is
evicted instead of throwing on every future eviction. Sweeping a symlink unlinks
it rather than following it.

## Tests

`src/test/local-cache.test.ts`, unit-testing `LocalCache` with explicit mtimes so
ordering doesn't depend on timestamp resolution:

| case | asserts |
| --- | --- |
| under limit | nothing evicted |
| over limit | trimmed to limit |
| `get()` on older entry | it survives, newer one evicted (LRU, not FIFO) |
| `markEntryRecentlyUsed()` on older entry | same, for the fresh path |
| `markEntryRecentlyUsed()` with no entry | no-op, no throw |
| older entry with a future mtime | the just-written entry is not the one evicted |
| `infinity` | nothing evicted |
| post-eviction | survivor still restorable |
| evicted entry | `get()` returns `undefined` |
| broken symlink in folder | eviction not wedged |
| stray file in folder | evicted like an entry |
| evicted entry | waits in trash, swept on demand, path shorter than its source |
| trash path is a file | cache stays over limit, `set()` still succeeds |
| two caches sweeping at once | no error |
| symlinked entry swept | the link goes, its target does not |
| aborted sweep | trash left for the next run |

Plus an end-to-end CLI test — 5 runs under `WIREIT_CACHE_MAX_ENTRIES=2`, 2
entries left and no trash behind — and a `cli-options` parse test. All run under
`npm run test:cache-local`.

## Compatibility

Previously unbounded caches are now trimmed; set
`WIREIT_CACHE_MAX_ENTRIES=infinity` to opt out. Existing over-limit folders trim
on that script's next cache write.

`.wireit/trash` is new, and transient — it is emptied at the end of the run that
creates it, or by the next run if that one is interrupted.


src/caching/local-cache.ts
```diff
@@ -58,13 +75,49 @@ export class LocalCache implements Cache {
       throw new Error(`Did not expect ${absCacheDir} to already exist.`);
     }
     await copyEntries(absoluteFiles, script.packageDir, absCacheDir);
+    await this.#evictLeastRecentlyUsedEntries(script);
```
aomarks: Let's make this more asynchronous. We don't need to block `set` on the cache eviction completing; it can run in the background while the rest of the run continues.

Here's a sketch of how we might want to do that:

Expose a `collectGarbage` method on the cache interface, instead of doing it automatically in `set`. The caller is responsible for calling it.

This method returns a Promise, and takes an abort signal. The abort signal would fire if the user wants to quit wireit early (e.g. sigint). We already have some plumbing for that.

If at the end of a wireit run (i.e. we would otherwise have exited), the GC promise has not resolved, we print a log a message to the user so that they know what they're waiting for, and mention it is safe to Ctrl-C to cancel. If the user cancels, the next run should resume the cleanup.

src/caching/local-cache.ts
```diff
@@ -58,13 +75,49 @@ export class LocalCache implements Cache {
...
+      // throwing on every future eviction.
+      const byRecency = await Promise.all(
+        entries.map(async (entry) => {
+          const path = pathlib.join(cacheDir, entry.name);
+          return {path, mtimeMs: (await fs.lstat(path)).mtimeMs};
+        }),
+      );
+      byRecency.sort((a, b) => a.mtimeMs - b.mtimeMs);
+      const doomed = byRecency.slice(0, byRecency.length - this.#maxEntries);
+      await Promise.all(
```
aomarks: To ensure we don't leave partially-deleted cache entry folders around which could cause future corrupted cache restores (e.g. if the user types Ctrl-C before we are done), let's first atomically move the doomed cache entry to a new trash folder (e.g. `.wireit/trash/<name>`, and then start deleting it. Then at wireit startup, we could check for anything sitting around in trash, and resume the deletions.

src/caching/local-cache.ts
aomarks: Let's make the cache eviction a bit more asynchronous and resilient to early exit.

Instead of blocking on the full delete completing here in `set`, let's only block on moving the doomed entries into a new `.wireit/trash/<name>-<random suffix>` folder (the random suffix prevents collisions in case we find the same cache entry >1 times before cleanup completes).

Then, expose a new `sweepTrash` method on the cache interface which tries to delete everything in the `trash` folder. It would return a promise and take an abort signal for early exit. Note we should also tolerate ENOENT, because two separate wireit processes in theory could be sweeping the same trash.

If the sweep promise hasn't resolved when the run would otherwise exit, we can log what we're waiting for and tell the user that Ctrl-C is safe. Anything left in trash will get picked up on the next run.

src/caching/local-cache.ts
```diff
@@ -34,6 +47,15 @@ export class LocalCache implements Cache {
       }
       throw error;
     }
+    // Recency lives in the mtime, so there is no index file to maintain. atime
+    // won't do, because filesystems are commonly mounted noatime or relatime.
+    const now = new Date();
+    try {
+      await fs.utimes(cacheDir, now, now);
```
aomarks: We should also bump the mtimes for the "fresh" case (which is where we don't actually need to restore anything from cache, because the output is already in place). Otherwise "fresh" hits won't count towards the cache entrie's recency score.

Note this can't happen in get, because on the fresh path we return before get is called (see the fresh check in `standard.ts`). Maybe we need a new touch(script, fingerprint) method on the cache interface.

src/cli-options.ts
```diff
@@ -46,6 +46,14 @@ export const packageDir = await (async (): Promise<string | undefined> => {
   }
 })();
 
+/**
+ * Deliberately small, because an entry is a full copy of a script's output, and
+ * scripts with large outputs are what make an unbounded cache painful. Two
+ * still covers what local caching is most useful for: bouncing between the
+ * current state and one other, like a branch you keep switching back to.
+ */
+const DEFAULT_CACHE_MAX_ENTRIES = 2;
```
aomarks: I think we could bump this up to maybe 10, so that the cache can survive a bit more churn when e.g. jumping between branches.

src/caching/local-cache.ts
```diff
@@ -58,13 +75,49 @@ export class LocalCache implements Cache {
...
+      }
+      // lstat, so a broken symlink in the folder gets evicted rather than
+      // throwing on every future eviction.
+      const byRecency = await Promise.all(
+        entries.map(async (entry) => {
+          const path = pathlib.join(cacheDir, entry.name);
+          return {path, mtimeMs: (await fs.lstat(path)).mtimeMs};
+        }),
+      );
+      byRecency.sort((a, b) => a.mtimeMs - b.mtimeMs);
```
aomarks, line 139: In theory (e.g. with coarse mtime and a low max entries), this could end up deleting a cache entry that we just wrote. Let's pass in any just-created entry and exclude it from the candidates so that this can't happen.

src/caching/local-cache.ts
```diff
@@ -18,9 +18,22 @@ import type {AbsoluteEntry} from '../util/glob.js';
 
 /**
  * Caches script output to each package's
- * ".wireit/<script-name-hex>/cache/<cache-key-sha256-hex>" folder.
+ * ".wireit/<script-name-hex>/cache/<cache-key-sha256-hex>" folder, keeping only
+ * the {@link maxEntries} least recently read or written entries per script.
```
aomarks: I think this should say "most recently read", not "least recently read".

aomarks (changes requested):
Thank you for the PR! I think this is a good approach overall, with a few comments around trash collection.

src/cli-options.ts
asmyshlyaev177: Ok, makes sense. Problem that I hit locally that whole node_modules were in cache, so better to have it configurable. Possibly that I just didn't read docs good enough.

aomarks:
Sorry for the very pedantic format check!

asmyshlyaev177:
> Sorry for the very pedantic format check!

It's ok, probably need to adjust it for markdown, something isn't right.

aomarks:
> > Sorry for the very pedantic format check!
> 
> It's ok, probably need to adjust it for markdown, something isn't right.

does it pass locally with `npm run lint`? do you have the same exact prettier version as the checked-in package lock dictates? `npm ci && npm run format:fix` should work, and if it doesn't there must be something wrong with our configuration.

asmyshlyaev177:
> > > Sorry for the very pedantic format check!
> > 
> > 
> > It's ok, probably need to adjust it for markdown, something isn't right.
> 
> does it pass locally with `npm run lint`? do you have the same exact prettier version as the checked-in package lock dictates? `npm ci && npm run format:fix` should work, and if it doesn't there must be something wrong with our configuration.

It does, pushed the fix.

asmyshlyaev177:
Somehow local prettier version drifted, just reinstalled it.

aomarks (approved):

aomarks:
Great, merged! Thanks for the PR. I have a few other small tasks to do before the next release, will post here when the release is out, probably in a couple days.

